PR #9, opened 2018-01-19
Develop

rmothilal:
Fixing unit tests

Stubbing SMTP server


rmothilal:
I have added more tests for settlements and the CSV process.

PR #54, opened 2018-03-21
Feature/#76 imp poc stream process rajiv

mdebarros:


API.md
```diff
@@ -20,6 +20,7 @@ The central ledger has two APIs targeted at different consumers. The DFSP API is
 * `POST` [**Get charge quote**](#get-a-charge-quote) 
 * `POST` [**Get authentication token**](#get-authentication-token)
 * `GET`  [**Health**](#health)
+* `GET`  [**Kafka**](#kafka)
```
mdebarros: @lazolalucas should this not be POST?

API.md
```diff
@@ -20,6 +20,7 @@ The central ledger has two APIs targeted at different consumers. The DFSP API is
...
 
 #### [Admin API](#admin-api) endpoints
```
mdebarros: This is not for the ADMIN API only: remember that the Admin API run its own separated thread from the normal API. We need potentially both (but at this moment only care about the normal API) to be able to have some kind of management capability going forward for the Kafka consumers. The APIs are for different things. Admin API is for the administrating of the Ledger, where the normal API is for the client usage of the Ledger. 

The Kafka API that I am asking for is more of a backend service only for maintenance purposes only, and potentially impacts both.

API.md
```diff
@@ -980,18 +981,29 @@ Get the current status of the service
...
 
 ##### Response
 ``` http
-HTTP/1.1 200 OK
+HTTP/1.1 201 OK
 {
-  "status": "OK"
-}
-```
+    "accountTopic": "topic-dfsp10-prepare-notification"
```
mdebarros: What is an "accountTopic"?

As per our discussion you need to also include the purpose of the consumer either in the URI or in the Body of the POST message. 

How will the Kafka consumer know if this should be a consumer for Notifications or Prepare flow?

src/domain/transfer/commands/index.js
```diff
@@ -43,6 +43,46 @@ const prepare = async (message) => {
   })
 }
 
+const debitNotificationPromise = (topicForDebitNotifications, notificationMsgForDebits, id) => {
```
mdebarros: @rmothilal: Why is there two different methods for the same notifications?

src/domain/transfer/commands/index.js
```diff
@@ -90,22 +130,60 @@ const prepareExecute = (payload, done) => {
...
+              Logger.info('result credit true')
+              return resolve(true)
+            } else {
+              done()
+              Logger.info('result credit false')
+              return reject(false)
+            }
+          })
+        }) */
+        return Promise.all([debitNotificationPromise(topicForDebitNotifications, notificationMsgForDebits, id), creditNotificationPromise(topicForCreditNotifications, notificationMsgForCredits, id)]).then(result => {
```
mdebarros: I see no difference between the debitNotificationPromise and the creditNotificationPromise methods except for the error message. Rather use a single method and add the error message as an input.

API.md
```diff
@@ -980,18 +981,29 @@ Get the current status of the service
...
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| status | String | A kafka consumer has been registered for your DFSP Account  |
+
+
 ##### Request
 ``` http
-GET http://central-ledger/health HTTP/1.1
+POST http://central-ledger/kafka HTTP/1.1
 ```
```
mdebarros: Where is the Post body?

PR #60, opened 2018-04-25
Feature/#149re align db model migrations

vgenev:


rmothilal:
Force merging branch for other members to continue with their stories

PR #64, opened 2018-04-27
Feature/#210refactor transfer batch

HenkKodde:
#210 - Changing transferBatch to transferSettlementBatch V1.0.0.

### DIFF
--- a/migrations/20180418101700_create-transfer-settlement-batch.js
+++ b/migrations/20180418101700_create-transfer-settlement-batch.js
@@ -1,13 +1,13 @@
 'use strict'
 
 exports.up = function(knex, Promise) {
-    return knex.schema.createTableIfNotExists('transferBatch', (t) => {
-        t.bigIncrements('transferBatchId').primary().notNullable()
+    return knex.schema.createTableIfNotExists('transferSettlementBatch', (t) => {
+        t.bigIncrements('transferSettlementBatchId').primary().notNullable()
         t.dateTime('createdDate').defaultTo(knex.fn.now()).notNullable()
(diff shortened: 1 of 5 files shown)


rmothilal:
Henk. You request has conflicts that need to be resolved. Also you have committed your change to disable the sidecar. Could you please pull in all changes from develop-PI2 from mojaloop into your branch and see if this resolves you conflicts. Also be careful of the files that you commit

config/default.json
```diff
@@ -11,7 +11,7 @@
     "SCALE": 2
   },
   "SIDECAR": {
-    "DISABLED": false,
```
rmothilal: This needs to remain false

src/shared/setup.js
```diff
@@ -58,9 +58,9 @@ const initialize = async function ({service, port, modules = [], runMigrations =
   await connectDatabase()
   await Sidecar.connect(service)
   const server = await createServer(port, modules)
-  if (service === 'api') {
+  /* if (service === 'api') {
     await Account.createLedgerAccount(Config.LEDGER_ACCOUNT_NAME, Config.LEDGER_ACCOUNT_PASSWORD, Config.LEDGER_ACCOUNT_EMAIL)
```
rmothilal: Now that this is disabled, are we not going to create a default account/participant?

vgenev: Yes, but it was disabled only to pass the 149 story without throwing errors. I will take care of it Monday

vgenev: Yes, but it was disabled only to pass the 149 story without throwing errors. I will take care of it Monday

PR #75, opened 2018-06-04
Feature/#291 align database diagram to latest implementation

ggrg:
The foreign key from transferStateChange to transfer was missing and current PR is to repair it.

As part of the reviewing process, please also have a look at the produced diagram:
central-ledger-schema-7


migrations/20180418101950_create-transfer-state-change.js
```diff
@@ -5,11 +5,8 @@ exports.up = function (knex, Promise) {
     t.bigIncrements('transferStateChangeId').primary().notNullable()
 
     t.string('transferId', 36).notNullable()
+    t.foreign('transferId').references('transferId').inTable('transfer')
```
mdebarros: Why has this been added as per references the commented out code below?

ggrg: They were renamed because the previous order did not work for adding a reference to the transfer table. The commented out code below actually has not been implemented in other migration script.

HenkKodde:
There is a lot of dead code (commented out) code in the Transfer Table. Is there a reason for it to be used at a later stage, or can we clean it up?

HenkKodde:
Just a note - we would also expect appropriate comments on the CDB to included going forward. I have a fair amount of work that was done for the quote, so will update the transfer state table with the following ".comment('Foreign key linking the transferState to the Transfer table.')".

PR #106, opened 2018-08-02
Feature/#306 #308 merged

vgenev:


vgenev:
two tests are left commented and few needs to be prepared later

PR #129, opened 2018-09-01
Several hot fixes across the board for v3.5.0-snapshot

mdebarros:
## Changes
1. Fulfil Error Handling Logic added by @shashi165
2. Issue with the Fulfil Condition due to a bug that never actually validated it correctly fixed by @shashi165
3. Re-aligned the Fulfil Condition Validation functions to the ML API v1.0 Spec
4. Added additional errors to the lib/errors.js enum
5. Updated errors in Fulfil Handler to use the lib/errors.js enum
6. Updated version to v3.5.0-snapshot

## Unit Test Results:
```
=============================== Coverage summary ===============================
Statements   : 68.13% ( 2157/3166 )
Branches     : 61.36% ( 459/748 )
Functions    : 50.93% ( 274/538 )
Lines        : 68.6% ( 2143/3124 )
================================================================================

1..755
# tests 755
# pass  755

# ok
```

## Integration Tests:
<details><summary>View Results</summary>
<p>

```XML
<?xml version="1.0"?>
<testsuites>
  <testsuite tests="0" failures="0" errors="0" name="Participant service"/>
  <testsuite tests="1" failures="0" errors="0" name="setup">
    <testcase name="#1 setup OK"/>
  </testsuite>
  <testsuite tests="0" failures="0" errors="0" name="create participant"/>
  <testsuite tests="0" failures="0" errors="0" name="getByName"/>
  <testsuite tests="0" failures="0" errors="0" name="getAll"/>
  <testsuite tests="0" failures="0" errors="0" name="Testing with participant"/>
  <testsuite tests="0" failures="0" errors="0" name="{"/>
  <testsuite tests="0" failures="0" errors="0" name="&quot;participantId&quot;: 1,"/>
  <testsuite tests="0" failures="0" errors="0" name="&quot;name&quot;: &quot;fsp11869180&quot;,"/>
  <testsuite tests="0" failures="0" errors="0" name="&quot;description&quot;: null,"/>
  <testsuite tests="0" failures="0" errors="0" name="&quot;isActive&quot;: 1,"/>
  <testsuite tests="0" failures="0" errors="0" name="&quot;createdDate&quot;: &quot;2018-09-01T00:31:09.000Z&quot;,"/>
  <testsuite tests="0" failures="0" errors="0" name="&quot;createdBy&quot;: &quot;unknown&quot;,"/>
  <testsuite tests="0" failures="0" errors="0" name="&quot;currencyList&quot;: ["/>
[...]

PR #132, opened 2018-09-04
Reverting Fulfilment Condition validation change until it can be discussed in PI3-packaging sprint

mdebarros:
Revert based on the following comment: https://github.com/mojaloop/central-ledger/pull/129/files/efa37caf16054555407f4a4f38c2acfa0d50a216#r214499966

Reverting to option 2 as per the code of PI2 which was never actually executed due to the lack of an error path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "3.5.1-snapshot",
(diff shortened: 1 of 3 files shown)


mdebarros:
Closing PR for now. Will re-open if it is required.

PR #133, opened 2018-09-05
Feature/#437 settlement window initialization

HenkKodde:
Updated 'settlement_3_settlementWindow.is' not to create a new settlementWindow if one exist in an open state. Story #437.


HenkKodde:
.

PR #143, opened 2018-09-21
fixes for bug 471, 472, 469

rmothilal:
this is for the ffg: 
- failing consumer when not found
- validation of fsp with better error messages
- committed fulfilment updating payeeFsp position

### DIFF
--- a/src/handlers/lib/kafka/consumer.js
+++ b/src/handlers/lib/kafka/consumer.js
@@ -95,7 +95,7 @@ const getConsumer = (topicName) => {
   if (listOfConsumers[topicName]) {
     return listOfConsumers[topicName].consumer
   } else {
-    return undefined
+    throw Error(`no consumer found for topic ${topicName}`)
   }
 }
 
--- a/src/handlers/lib/kafka/consumer.js
+++ b/src/handlers/lib/kafka/consumer.js
@@ -95,7 +95,7 @@ const getConsumer = (topicName) => {
   if (listOfConsumers[topicName]) {
     return listOfConsumers[topicName].consumer
   } else {
-    throw Error(`no consumer found for topic ${topicName}`)
+    return undefined
   }
 }
 
--- a/src/handlers/lib/kafka/consumer.js
+++ b/src/handlers/lib/kafka/consumer.js
@@ -95,7 +95,7 @@ const getConsumer = (topicName) => {
   if (listOfConsumers[topicName]) {
     return listOfConsumers[topicName].consumer
   } else {
-    throw Error(`no consumer found for topic ${topicName}`)
+    throw Error(`No consumer found for topic ${topicName}`)
   }
 }
 
@@ -113,7 +113,7 @@ const isConsumerAutoCommitEnabled = (topicName) => {
   if (listOfConsumers[topicName]) {
     return listOfConsumers[topicName].autoCommitEnabled
   } else {
-    throw Error(`no consumer found for topic ${topicName}`)
+    throw Error(`No consumer found for topic ${topicName}`)
   }
 }
 
--- a/src/handlers/positions/handler.js
+++ b/src/handlers/positions/handler.js
@@ -86,9 +86,13 @@ const positions = async (error, messages) => {
       Logger.info('PositionHandler::positions::prepare')
       kafkaTopic = Utility.transformAccountToTopicName(message.value.from, TransferEventType.POSITION, TransferEventAction.PREPARE)
       consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-      const { preparedMessagesList, limitAlarms } = await PositionService.calculatePreparePositionsBatch(prepareBatch)
+      if (!consumer) {
+        Logger.info(`no consumer found for topic ${kafkaTopic}`)
+        return true
+      }
+      const {preparedMessagesList, limitAlarms} = await PositionService.calculatePreparePositionsBatch(prepareBatch)
       for (let prepareMessage of preparedMessagesList) {
-        const { transferState, rawMessage } = prepareMessage
+        const {transferState, rawMessage} = prepareMessage
         if (transferState.transferStateId === Enum.TransferState.RESERVED) {
           await Utility.produceGeneralMessage(TransferEventType.TRANSFER, TransferEventAction.PREPARE, rawMessage.value, Utility.ENUMS.STATE.SUCCESS)
         } else {
@@ -107,7 +111,10 @@ const positions = async (error, messages) => {
       Logger.info('PositionHandler::positions::commit')
       kafkaTopic = Utility.transformAccountToTopicName(message.value.from, TransferEventType.POSITION, TransferEventType.FULFIL)
       consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-
+      if (!consumer) {
+        Logger.info(`no consumer found for topic ${kafkaTopic}`)
+        return true
+      }
       // Check current transfer state
       const transferInfo = await TransferService.getTransferInfoToChangePosition(payload.transferId, Enum.TransferParticipantRoleType.PAYEE_DFSP, Enum.LedgerEntryType.PRINCIPLE_VALUE)
       if (transferInfo.transferStateId !== TransferState.RECEIVED_FULFIL) {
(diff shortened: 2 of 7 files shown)


rmothilal:
@mdebarros it issue is that we cannot commit the message to the consumer because it does not exist anymore. Do you suggest we recreate the consumer and commit the message? I attempted to commit the message but that failed.

src/models/transfer/facade.js
```diff
@@ -273,7 +277,7 @@ const saveTransferPrepared = async (payload, stateReason = null, hasPassedValida
       value: payload.ilpPacket
     }
 
-    const state = ((hasPassedValidation) ? Enum.TransferState.RECEIVED_PREPARE : Enum.TransferState.INVALID)
+    const state = ((hasPassedValidation) ? Enum.TransferState.RECEIVED_PREPARE : Enum.TransferState.ABORTED)
```
elnyry: Regarding this one - I think the intention was to use INVALID internally to easily identify requests for which validation failed. Here is the sequence diagram - https://github.com/mojaloop/docs/blob/develop/CentralServices/seq_diagrams/seq-prepare-1.1.1.a.svg

rmothilal: @elnyry Invalid does not exist in the transferState table. This will fail because of a foreign key exception

elnyry: Its in the seeds here - https://github.com/mojaloop/central-ledger/blob/develop/seeds/transferState.js ?

mdebarros:
The issue is not that we require a new consumer, it's that we are receiving a message in the wrong topic i.e. consumer prepare for dfsp1 , that consumer the tries to commit the message to the right topic which is for dfsp2.

The correct approach is that we should be logging the error, and then dfsp1 consumer should commit the message.

I.e. the way we look up the consumer is incorrect. We should be looking to the consumer not on the message.value (payload) but rather on the message meta data which hopefully contains the topic from where the message was attained from.

rmothilal:
rework to still throw error for missing topic and handle appropriately as well as getting topic from incoming message and not creating it from the fsp. reverted state to invalid for validation errors on prepare. fixed corresponding tests

src/handlers/positions/handler.js
```diff
@@ -204,10 +210,12 @@ const positions = async (error, messages) => {
...
       Logger.info('PositionHandler::positions::invalidEventTypeOrAction')
-      kafkaTopic = Utility.transformAccountToTopicName(message.value.from, message.value.metadata.event.type, message.value.metadata.event.action)
-      consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-      if (!consumer) {
+      kafkaTopic = message.topic
+      try {
+        consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+      } catch (e) {
         Logger.info(`no consumer found for topic ${kafkaTopic}`)
+        Logger.info(e)
```
mdebarros: Should be a loger.error.

src/handlers/positions/handler.js
```diff
@@ -181,10 +185,12 @@ const positions = async (error, messages) => {
...
         newMessage.value.content.payload = Utility.createPrepareErrorStatus(3303, reason, newMessage.value.content.payload.extensionList)
-        kafkaTopic = Utility.transformAccountToTopicName(newMessage.value.from, TransferEventType.POSITION, TransferEventAction.ABORT)
-        consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-        if (!consumer) {
+        kafkaTopic = message.topic
+        try {
+          consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+        } catch (e) {
           Logger.info(`no consumer found for topic ${kafkaTopic}`)
+          Logger.info(e)
```
mdebarros: Should be a loger.error.

src/handlers/positions/handler.js
```diff
@@ -137,10 +139,12 @@ const positions = async (error, messages) => {
...
       Logger.info('PositionHandler::positions::reject')
-      kafkaTopic = Utility.transformAccountToTopicName(message.value.from, TransferEventType.POSITION, TransferEventAction.ABORT)
-      consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-      if (!consumer) {
+      kafkaTopic = message.topic
+      try {
+        consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+      } catch (e) {
         Logger.info(`no consumer found for topic ${kafkaTopic}`)
+        Logger.info(e)
```
mdebarros: Should be a loger.error.

src/handlers/positions/handler.js
```diff
@@ -84,10 +84,12 @@ const positions = async (error, messages) => {
...
       Logger.info('PositionHandler::positions::prepare')
-      kafkaTopic = Utility.transformAccountToTopicName(message.value.from, TransferEventType.POSITION, TransferEventAction.PREPARE)
-      consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-      if (!consumer) {
+      kafkaTopic = message.topic
+      try {
+        consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+      } catch (e) {
         Logger.info(`no consumer found for topic ${kafkaTopic}`)
+        Logger.info(e)
```
mdebarros: Should be a loger.error.

src/handlers/lib/kafka/consumer.js
```diff
@@ -95,7 +95,7 @@ const getConsumer = (topicName) => {
   if (listOfConsumers[topicName]) {
     return listOfConsumers[topicName].consumer
   } else {
-    return undefined
+    throw Error(`no consumer found for topic ${topicName}`)
```
mdebarros: Should we not capitalise the message?  -> "No consumer....."

src/handlers/transfers/handler.js
```diff
@@ -95,11 +95,14 @@ const prepare = async (error, messages) => {
...
     Logger.info('TransferService::prepare')
-    const kafkaTopic = Utility.transformAccountToTopicName(message.value.from, TransferEventType.TRANSFER, TransferEventAction.PREPARE)
-    const consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-    if (!consumer) {
+    const kafkaTopic = message.topic
+    try {
+      consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+    } catch (e) {
       Logger.info(`no consumer found for topic ${kafkaTopic}`)
+      Logger.info(e)
```
mdebarros: Should be a loger.error.

src/handlers/transfers/handler.js
```diff
@@ -228,10 +231,13 @@ const fulfil = async (error, messages) => {
...
-    const kafkaTopic = Utility.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventType.FULFIL)
-    const consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-    if (!consumer) {
+    const kafkaTopic = message.topic
+    let consumer
+    try {
+      consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+    } catch (e) {
       Logger.info(`no consumer found for topic ${kafkaTopic}`)
+      Logger.info(e)
```
mdebarros: Should be a loger.error.

src/handlers/transfers/handler.js
```diff
@@ -342,13 +348,15 @@ const transfer = async (error, messages) => {
...
     if (action.toLowerCase() === TransferEventAction.PREPARE && status.toLowerCase() === TransferEventStatus.SUCCESS) {
-      const kafkaTopic = Utility.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.TRANSFER)
-      const consumer = Kafka.Consumer.getConsumer(kafkaTopic)
-      if (!consumer) {
+      const kafkaTopic = message.topic
+      try {
+        consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+      } catch (e) {
         Logger.info(`no consumer found for topic ${kafkaTopic}`)
+        Logger.info(e)
```
mdebarros: Should be a loger.error.

src/handlers/transfers/handler.js
```diff
@@ -359,8 +367,14 @@ const transfer = async (error, messages) => {
...
       return true
     } else if (action.toLowerCase() === TransferEventAction.COMMIT && status.toLowerCase() === TransferEventStatus.SUCCESS) {
-      const kafkaTopic = Utility.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.TRANSFER)
-      const consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+      const kafkaTopic = message.topic
+      try {
+        consumer = Kafka.Consumer.getConsumer(kafkaTopic)
+      } catch (e) {
+        Logger.info(`no consumer found for topic ${kafkaTopic}`)
+        Logger.info()
```
mdebarros: Should be a loger.error.

PR #149, opened 2018-09-28
Updated populateTestData script to set endPoint Callback URLs for each Participant, and removed instructions from the restartMockServer.sh to configure the EndPoints in the ML-API-Adapter config which is no longer necessary.

mdebarros:


elnyry:
Merging this as Miguel is away

PR #235, opened 2019-01-23
Feature/585 Logging participants and accounts activation and deactivation

ggrg:
[Story #585](https://github.com/mojaloop/project/issues/585)


src/admin/participants/handler.js
```diff
@@ -161,6 +162,11 @@ const update = async function (request) {
   Sidecar.logRequest(request)
   try {
     const updatedEntity = await Participant.update(request.params.name, request.payload)
+    if (request.payload.isActive !== undefined) {
+      const isActiveText = request.payload.isActive ? 'activated' : 'disabled'
```
rmothilal: Should this be hardcoded or put into the enum?

src/admin/participants/handler.js
```diff
@@ -312,6 +318,11 @@ const updateAccount = async function (request, h) {
       ledgerAccountType: await request.server.methods.enums('ledgerAccountType')
     }
     await Participant.updateAccount(request.payload, request.params, enums)
+    if (request.payload.isActive !== undefined) {
+      const isActiveText = request.payload.isActive ? 'activated' : 'disabled'
```
rmothilal: Should this be hardcoded or pout into the enum?

src/admin/participants/handler.js
ggrg: I see your point and I've created local enum object. I don't think we should flood the global Enum with debugging messages, but probably have a story in future where we isolate all debug messages into a single file, which would later ease i18n. @elnyry, please backlog if appropriate.

src/admin/participants/handler.js
ggrg: same as previous

PR #236, opened 2019-01-24
added new message to notification topic for limit adjustment.

vgenev:
Added producing new message to the Kafka Notification Topic for limit adjustment through Admin API


src/domain/participant/index.js
```diff
@@ -447,16 +447,53 @@ const getLimitsForAllParticipants = async ({ currency = null, type = null }) =>
  * @returns {integer} - Returns number of database rows affected if successful, or throws an error if failed
  */
 
+// {
```
ggrg: Please, avoid adding commented out code. Either update the example above, that is part of jsdoc or add another example, if different

src/domain/participant/index.js
```diff
@@ -447,16 +447,53 @@ const getLimitsForAllParticipants = async ({ currency = null, type = null }) =>
...
+//       "value": 1500,
+//       "thresholdAlarmPercentage": 10
+//   }
+// }
+
 const adjustLimits = async (name, payload) => {
   try {
-    const participant = await ParticipantFacade.getByNameAndCurrency(name, payload.currency, Enum.LedgerAccountType.POSITION)
+    let { limit, currency } = payload
+    payload.name = name
```
ggrg: I'd move this line just before line 466, where payload is used, not because it is bad in the beginning, but on the previous line you have payload object decomposing, and then you add other properties.

PR #238, opened 2019-01-28
Аdded limit-adjustment action to the action map for topic naming, and message for initial limit

vgenev:


src/domain/participant/index.js
```diff
@@ -278,6 +278,8 @@ const addLimitAndInitialPosition = async (participantName, limitAndInitialPositi
     if (limitAndInitialPosition.initialPosition == null) {
       limitAndInitialPosition.initialPosition = Config.PARTICIPANT_INITIAL_POSTITION
     }
+    let payload = Object.assign({}, limitAndInitialPositionObj, {name: participantName})
+    await Utility.produceGeneralMessage(TransferEventType.NOTIFICATION, 'limit-adjustment', createLimitAdjustmentMessageProtocol(payload), Utility.ENUMS.STATE.SUCCESS)
```
ggrg: Please add enum and refer to it instead of using a string within the code here and also on the following lines: 460, 467. I would go with either limitEventAction.LIMIT_ADJUSTMENT or eventAction.LIMIT_ADJUSTMENT.

src/domain/participant/index.js
```diff
@@ -455,18 +457,18 @@ const adjustLimits = async (name, payload) => {
...
   }
 }
 
-const createLimitAdjustmentMessageProtocol = (payload, action = 'limitAdjustment', state = '', pp = '') => {
+const createLimitAdjustmentMessageProtocol = (payload, action = 'limit-adjustment', state = '', pp = '') => {
   return {
     id: Uuid(),
     from: payload.name,
-    to: 'SYSTEM',
+    to: Config.HUB_NAME,
```
ggrg: If we have config Config.HUB.NAME in central-settlement, probably it's good idea to unify it. [More info](https://github.com/mojaloop/central-settlement/pull/58).

src/domain/participant/index.js
vgenev: done

src/domain/participant/index.js
vgenev: This is due to the use of RC module. I can delete the RC values in central-settlement and it will be the same on all places, including CEP.

ggrg:
@vgenev, please commit your changes, if you're done.

PR #240, opened 2019-01-31
Update Onboarding.md

kakoma:
Add path to valid Postman collection to import during setup


kakoma:
Thanks @kjw000! I'll keep adding as I get my local setup up-to-speed.

PR #246, opened 2019-02-12
Hotfix/Transfer Position Topic

ggrg:


ggrg:
This PR is no longer need and has been addressed with https://github.com/mojaloop/central-ledger/pull/247

PR #247, opened 2019-02-13
bugfix/625 Fix Growl Vulnerability

ggrg:
* Fixed Growl 1.10.5 vulnerability by removing devDependency on five-bells-ledger-api-tests which is currently not used at all
* Enabled handlers.test.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,6 @@
   },
   "devDependencies": {
     "faucet": "0.0.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "5.1.2",
+  "version": "5.1.3",
```
elnyry: Looks like it was already bumped up in another PR, 5.1.2 should be good until we do another release

ggrg: done

PR #257, opened 2019-03-13
Bugfix/688 FundsIn for inactive participant

ggrg:
- Issue fixed
- Corrected unit tests accordingly
- Achieved full coverage for unit tests

### DIFF
--- a/src/domain/participant/index.js
+++ b/src/domain/participant/index.js
@@ -49,7 +49,6 @@ const AccountNotFoundErrorText = 'Account not found'
 const AccountNotPositionTypeErrorText = 'Only position account update is permitted'
 const AccountNotSettlementTypeErrorText = 'Account is not SETTLEMENT type'
 const ActionNotSupportedText = 'The action is not supported'
-const CreateParticipantGenericErrorText = 'Something went wrong. Participant cannot be created'
 const ParticipantAccountCurrencyMismatchText = 'The account does not match participant or currency specified'
 const ParticipantAccountMismatchText = 'Participant/account mismatch'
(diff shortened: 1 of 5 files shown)


src/domain/participant/index.js
```diff
@@ -43,10 +43,23 @@ const Enum = require('../../lib/enum')
 const TransferEventType = Enum.transferEventType
 const TransferEventAction = Enum.transferEventAction
 
+// Alphabetically ordered list of error texts used below
+const AccountInactiveErrorText = 'Account is currently set inactive'
+const AccountNotFoundErrorText = 'Account not found'
+const AccountNotPositionTypeErrorText = 'Only position account update is permitted'
+const AccountNotSettlementTypeErrorText = 'Account is not SETTLEMENT type'
+const ActionNotSupportedText = 'The action is not supported'
+const CreateParticipantGenericErrorText = 'Something went wrong. Participant cannot be created'
```
mdebarros: Is this a valid error?

src/domain/participant/index.js
```diff
@@ -59,7 +58,6 @@ const ParticipantNotFoundText = 'Participant does not exist'
 const create = async (payload) => {
   try {
     const participant = await ParticipantModel.create({ name: payload.name })
-    if (!participant) throw new Error(CreateParticipantGenericErrorText)
```
mdebarros: As per our discussion. This scenario will never occur since the create as an insert and any errors will be handled by the `ParticipantModel`.

ggrg: true

PR #259, opened 2019-03-15
hotfix/Validate Header During Transfer Prepare

ggrg:
- Validate header during transfer prepare to make sure payer from the payload matches header data

### DIFF
--- a/src/handlers/transfers/validator.js
+++ b/src/handlers/transfers/validator.js
@@ -85,7 +85,16 @@ const validatePositionAccountByNameAndCurrency = async function (participantName
 const validateDifferentDfsp = (payload) => {
   const isPayerAndPayeeDifferent = (payload.payerFsp !== payload.payeeFsp)
   if (!isPayerAndPayeeDifferent) {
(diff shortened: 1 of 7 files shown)


src/handlers/transfers/validator.js
```diff
@@ -85,7 +85,16 @@ const validatePositionAccountByNameAndCurrency = async function (participantName
...
   if (!isPayerAndPayeeDifferent) {
-    reasons.push(`Payer and Payee should be different.`)
+    reasons.push(`Payer and Payee should be different`)
+    return false
+  }
+  return true
+}
+
+const validateFspiopSourceMatchesPayer = (payload, headers) => {
+  const matched = (headers && headers['fspiop-source'] && headers['fspiop-source'] === payload.payerFsp)
```
mdebarros: What about validating the payeeFsp for the Fulfils?

ggrg: I agree, but this "hotfix" update is only for the prepare.

PR #260, opened 2019-03-18
Hotfixes for POST-GET /transfer operations

mdebarros:
PR for https://github.com/mojaloop/project/issues/687 which includes fixes for:
- POST /transfers when fulfil transferState='ABORTED' now returns the correct payload
- GET /transfers when fulfil transferState='ABORTED' now returns the correct payload
- Minor bug in the Participant domain was not returning the headers as the message contained `header` instead of `headers` key
- GET /transfers was not returning the correct error when a transfer ID was not found. This now returns `ERROR 3208`.
- Central-switch designation has been changed to `switch` which is aligned to the Mojaloop specification

Maintenance changes:
- Added a new Util.clone() method and updated most (if not all) clone methods to use this common operation
- Moved  toFulfil transform method from the Transfer Handlers to transformer.js
- Added unit tests for the toFulfil transformer method to ensure code-coverage is maintained
- Added new ENUM for TransferStateEnum which is required for logical comparison within the Transfer Handlers
- Updated all Transfer Transformers to return only fields that have values, and leave out any undefined/null key-values from the output transformation result in transformer.js
- Bumped version to 5.4.0


mdebarros:
PR waiting for https://github.com/mojaloop/central-ledger/pull/259 before merge-conflicts can be resolved.

src/lib/enum.js
```diff
@@ -194,6 +194,13 @@ const TransferState = {
   RESERVED_TIMEOUT: 'RESERVED_TIMEOUT'
 }
 
+const TransferStateEnum = {
```
rmothilal: whats the difference between this and the TransferState constant above this? Should these no be added to that enum?

src/lib/util.js
```diff
@@ -56,6 +56,19 @@ const filterUndefined = (fields) => {
...
+/**
+ * Method to provide object clonning
+ *
+ * TODO:
+ *  Implement a better deep copy method
+ *
+ * @param value
+ * @returns {any}
+ */
+const clone = (value) => {
```
rmothilal: i think the name of this method is confusing because it actually creates a JSON object from a string, so not clone more transform

mdebarros: It is a clone, it is however only a shallowClone since its only copying properties and not functions. Do you not agree?

src/lib/enum.js
mdebarros: They are not the same.

`TransferState` is the internal state representation where `TransferStateEnum` is the "Mojaloop" specification enumerated states.

PR #264, opened 2019-03-31
Bugfix/697 698 Http Headers Validation

ggrg:


src/handlers/lib/utility.js
```diff
@@ -513,6 +513,7 @@ const proceed = async (params, opts) => {
   if (fromSwitch) {
     message.value.to = message.value.from
     message.value.from = Enum.headers.FSPIOP.SWITCH
+    delete message.value.content.headers.accept
```
mdebarros: Hi. This is handled l by the transformer in ml-api. Please remove it from here.

ggrg: fixed

PR #267, opened 2019-04-04
Feature/693 Bulk Transfers Design

ggrg:
- Bulk Transfers Migrations
- Bulk Transfer State lookup values
- Bulk Processing State lookup values


ggrg:
@mdebarros, it is just I preferred writing knex migration scripts and commit to github, rather than working with MySQL Workbench and having migration scripts work due.

ggrg:
Closing this PR as it is now outdates and migrations have been moved [here](https://github.com/ggrg/central-ledger/tree/feature/745bulkPrepareHandlerImpl)

PR #269, opened 2019-04-25
feature/742 String Type Error Codes

ggrg:
- errorCode.toString


src/models/transfer/transferError.js
```diff
@@ -106,7 +106,7 @@ const getByTransferStateChangeId = async (transferStateChangeId) => {
 
 const getByTransferId = async (id) => {
   try {
-    return await Db.transferError.query(async (builder) => {
+    let transferError = await Db.transferError.query(async (builder) => {
```
rmothilal: should the column for errorCode here not be altered to be a string value

PR #270, opened 2019-04-30
Feature/724Db Object Exposed By src/lib

ggrg:


ggrg:
It’s better for two reasons:
1. It would be easier to change `central-services-database` call in future in case we move away from it or change its structure (e.g. **.Db**)
2. See this stub: https://github.com/mojaloop/central-ledger/pull/270/files#diff-69253a9df5835623e4df79c23dae53bcR109 (+6 similar). Changing to what you have suggested would require some additional stubbing in tests, which might be tricky, as it is **Proxyquire**!

PR #277, opened 2019-05-20
762 optimize dockerfiles

lewisdaly:
Part of https://github.com/mojaloop/project/issues/762

- Reordered and optimized the commands for `Dockerfile`, `test-integration.Dockerfile` and `test.Dockerfile` to take better advantage of docker layer caching.
- Changed base image to official alpine image to `node:10.15.3-alpine`
- Added a `--cache-from` flag in the functional test runner to improve performance of running functional tests locally

All of these images build and test locally just fine.

### DIFF
--- a/test/integration-runner.sh
+++ b/test/integration-runner.sh
@@ -184,7 +184,7 @@ is_db_up() {
 stop_docker
 
 >&1 echo "Building Docker Image $DOCKER_IMAGE:$DOCKER_TAG with $DOCKER_FILE"
-docker build --no-cache -t $DOCKER_IMAGE:$DOCKER_TAG -f $DOCKER_FILE .
+docker build --cache-from $DOCKER_IMAGE:$DOCKER_TAG -t $DOCKER_IMAGE:$DOCKER_TAG -f $DOCKER_FILE .
 
 if [ "$?" != 0 ]
 then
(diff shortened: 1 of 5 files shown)


test/integration-runner.sh
```diff
@@ -184,7 +184,7 @@ is_db_up() {
 stop_docker
 
 >&1 echo "Building Docker Image $DOCKER_IMAGE:$DOCKER_TAG with $DOCKER_FILE"
-docker build --no-cache -t $DOCKER_IMAGE:$DOCKER_TAG -f $DOCKER_FILE .
+docker build --cache-from $DOCKER_IMAGE:$DOCKER_TAG -t $DOCKER_IMAGE:$DOCKER_TAG -f $DOCKER_FILE .
```
mdebarros: I am not 100% sure about the `--cache-from` since you are referring to the $DOCKER_TAG.

I only see this optimised a "localised" Integration Test where a local Docker registry is maintained. I do not see this being of any advantage on CircleCI since there is no previous build history.

What would occur if there is no pre-existing cache? I assume the build would ignore the cache and build from no-cache?

lewisdaly: Yep you're right, it will only speed up running the integration tests locally where the image had already been built. There is a way to enable docker image caching inside of CircleCI, but they charge an extra fee for this, and from what I saw the builds weren't too crazy slow anyway.

If there's no cache it just build a new docker image from scratch. The command won't fail.

PR #280, opened 2019-05-21
Feature/#769 additional party identifiers transaction scenarios

HenkKodde:
- Updated 'PartyIdentifiers' Seeds for quoting services.
- Updated 'TransactionSenarions' Seeds for quoting services.
- Updated release from 6.2.3 to 6.2.4

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -7,7 +7,7 @@
     "SCALE": 2
   },
   "SIDECAR": {
-    "DISABLED": "true",
+    "DISABLED": false,
     "HOST": "localhost",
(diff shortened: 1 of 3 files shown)


HenkKodde:
Have tested all the new scenarios and combinations. Verified everything is writing to the database and populated in the quote table in the transactionIndicatorTypeID and transactionScenarioId columns appropriately. Have also tested various combinations of the new additions. everything populated as expected.
Negative verifications was also perform, and they were not accepted - as per expectation.

PR #285, opened 2019-06-06
Feature/796 improved health check

lewisdaly:
Part of: mojaloop/project#796

Implement an improved health check for the `central-ledger`. 

This PR includes:
- created a new `handler.js` for root api endpoints
- implemented new health check 
- unit tests and integration tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "faucet": "0.0.1",
     "istanbul": "1.1.0-alpha.1",
     "jsonpath": "1.0.0",
-    "nodemon": "^1.19.1",
+    "nodemon": "1.19.1",
     "pre-commit": "1.2.2",
     "proxyquire": "2.1.0",
     "rewire": "4.0.1",
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "scripts": {
     "start": "run-p start:api",
     "start:api": "node src/api/index.js",
+    "watch:api": "nodemon src/api/index.js",
     "start:handlers": "node src/handlers/index.js",
     "dev": "npm run docker:stop && docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build -d",
     "test": "npm run test:unit | faucet",
@@ -109,6 +110,7 @@
     "faucet": "0.0.1",
     "istanbul": "1.1.0-alpha.1",
     "jsonpath": "1.0.0",
+    "nodemon": "^1.19.1",
     "pre-commit": "1.2.2",
     "proxyquire": "2.1.0",
     "rewire": "4.0.1",
--- a/src/api/root/handler.js
+++ b/src/api/root/handler.js
@@ -50,16 +50,8 @@ const serviceName = {
  * Gets the health status of the service.
  *
  *
- * if detailed=false, returns a basic health check:
- *
- * {
- *   "status": "OK",
- *   "uptime": 0,
- *   "started": "2019-05-31T05:09:25.409Z",
- *   "versionNumber": "5.2.3"
- * }
- *
- * if detailed=true, returns the health of related sub-services:
+ * @example:
+ * returns the health of related sub-services:
  * {
  *   "status": "OK",
  *   "uptime": 0,
@@ -67,7 +59,7 @@ const serviceName = {
  *   "versionNumber": "5.2.3",
  *   "services": [
  *     {
- *       "name": "kafka",
+ *       "name": "datastore",
  *       "status": "OK"
  *     }
  *   ]
@@ -144,11 +136,7 @@ const getSubServiceHealth = async function (serviceName) {
         Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.POSITION),
         Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.FULFIL)
       ]
-
-      // TODO: not sure if this is needed
-      const producerTopics = [
-        // 'topic-notification-event',
-      ]
+      const producerTopics = []
 
       let status = statusEnum.OK
       try {
@@ -176,8 +164,6 @@ const getSubServiceHealth = async function (serviceName) {
  * if any service.status is DOWN, then the entire service
  * is considered unhealthy (will return false)
  *
- * If a sub-service is DISABLED, and the other sub-services are OK, then
- * the entire service is considered healthy
  */
 const evaluateServiceHealth = function (services) {
   return services.reduce((acc, curr) => {
--- a/src/api/root/handler.js
+++ b/src/api/root/handler.js
@@ -136,12 +136,10 @@ const getSubServiceHealth = async function (serviceName) {
         Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.POSITION),
         Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.FULFIL)
       ]
-      const producerTopics = []
 
       let status = statusEnum.OK
       try {
         await P.all(consumerTopics.map(t => Kafka.Consumer.getConsumer(t)))
-        await P.all(producerTopics.map(t => Kafka.Producer.getProducer(t)))
       } catch (err) {
         status = statusEnum.DOWN
       }
--- a/src/api/root/handler.js
+++ b/src/api/root/handler.js
@@ -25,10 +25,11 @@
 const P = require('bluebird')
 
 const packageJson = require('../../../package.json')
-const Db = require('../../lib/db')
 const Util = require('../../handlers/lib/utility')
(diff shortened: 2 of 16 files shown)


mdebarros:
Note that the test-coverage failed. Let me know if you need any help with that.

lewisdaly:
@mdebarros coverage checks are passing now.

package.json
```diff
@@ -109,6 +110,7 @@
     "faucet": "0.0.1",
     "istanbul": "1.1.0-alpha.1",
     "jsonpath": "1.0.0",
+    "nodemon": "^1.19.1",
```
mdebarros: Please remove the carrot.

src/api/root/handler.js
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * getSubServiceHealth
+ *
+ * Gets the health status of a sub-service. Takes advantage of existing
+ * connections to the service (doesn't implement a ping or )
+ */
+const getSubServiceHealth = async function (serviceName) {
+  switch (serviceName) {
+    case 'datastore': {
+      const tables = await Db._listTables()
```
mdebarros: We should also consider the following approaches for the DB health checks:

1. Check the migration_lock table `select is_locked from migration_lock;`. If `is_locked` === `0` we are all good. This is the same check we use on the init-containers.
2 `SELECT version()`. This purely checks connectivity. This is what was used in the init-containers before we implemented the readiness check on 1. above.

The nice thing about option 1 is that it guarantees that the DB is in a good state for processing. However, the 2nd option should be ok since we are checking option 1 on the init-container anyway, and on the health end-point we are more concerned about the connectivity to the db.

Let me know your thoughts here.

src/api/root/handler.js
```diff
@@ -0,0 +1,183 @@
...
+
+      return {
+        name: serviceName,
+        status: tables.length > 0 ? statusEnum.OK : statusEnum.DOWN
```
mdebarros: This would potentially change if either option 1 or 2 approach is selected.

src/api/root/handler.js
```diff
@@ -0,0 +1,183 @@
...
+      */
+      const consumerTopics = [
+        Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.PREPARE),
+        Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.POSITION),
+        Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.FULFIL)
+      ]
+
+      let status = statusEnum.OK
+      try {
+        await P.all(consumerTopics.map(t => Kafka.Consumer.getConsumer(t)))
```
mdebarros: The Kafka streaming lib (https://github.com/mojaloop/central-services-stream) does not expose any method to check connectivity, however, the node-rdkafka library does with "isConnected()". It may be better to update the library appropriately to expose a similar method to support the health check purpose.

The approach above will work on the initial setup of the service, but we have seen times where the consumers are disconnected. This approach would still report "ok" in that scenario.

src/api/root/handler.js
lewisdaly: Yeah we can really go either way on this.

Is there a chance that in the future we might want to run migrations while the API servers are still alive for some reason (for example minimizing downtime between deployments)? In that case *1* might make more sense so that kubernetes takes the node out of service. 

The same thing goes for relying on init containers, perhaps if the init containers were to fail to unlock the migration lock (or some other random bug), and we relied on *2*, then we could get in a tricky situation where we are trying to connect to the database when we shouldn't be.

src/api/root/handler.js
lewisdaly: Sure. I'll take a look at updating that library.

mdebarros: > Sure. I'll take a look at updating that library.
> 
> We might also have to handle a case where there are different hosts for each of the topics as well - as I've noticed that the hostname of the broker can be set per topic in the config file.

Totally agree with you there, but implementing the "isConnected()" method should cater for this since you will grab all the consumers and just run the "isConnected" method to ensure that all is well.

src/api/root/handler.js
mdebarros: > Is there a chance that in the future we might want to run migrations while the API servers are still alive for some reason (for example minimizing downtime between deployments)? In that case _1_ might make more sense so that kubernetes takes the node out of service.

That is a valid point. That could occur on an upgrade through Helm and assuming that there were changes to the migration scripts....this would cause the migration to run. However, that may not be a bad thing as per your comment since you probably don't want to access the DB while the migration is running. Also, I wouldn't recommend upgrading with this approach if the system is still busy processing "live transactions". It would require some scheduled downtime.

> The same thing goes for relying on init containers, perhaps if the init containers were to fail to unlock the migration lock (or some other random bug), and we relied on _2_, then we could get in a tricky situation where we are trying to connect to the database when we shouldn't be.

You are 100% correct. But in theory the init-container is a gatekeeper, so the main container should never start up since the migration lock is never released. Thus Kubernetes will never have the opportunity to ever run any health checks against the main container.

rmothilal:
@lewisdaly @mdebarros Do we also not have health checkpoints on the kafka handlers?

mdebarros:
> @lewisdaly @mdebarros Do we also not have health checkpoints on the kafka handlers?
> Please have a look at src/handler/api/routes.js

Good point! @lewisdaly take note of the health end-points for the kafka handlers! Totally forgot about those.

lewisdaly:
@mdebarros @rmothilal I can't see any special kafka-handling health checks there, only the same as the other health api health checks:

`src/handlers/api/routes.js`
```js
module.exports = [
  {
    method: 'GET',
    path: '/health',
    handler: function (request, h) {
      return h.response({ status: 'OK' }).code(200)
    },
    options: {
      tags
    }
  }
]
```

lewisdaly:
@mdebarros regarding calling `isConnected()` on the `node-rdkafka` library; after doing some more digging around, I don't think that `isConnected()` does what the name implies.

According to [this](https://github.com/Blizzard/node-rdkafka/issues/217#issuecomment-313582908) thread, another person was making a health check for the library, and `isConnected` only returns false when you *manually* disconnect. The authors recommend instead to poll `getMetadata` for a better health check.

rmothilal:
> @mdebarros @rmothilal I can't see any special kafka-handling health checks there, only the same as the other health api health checks:
> 
> `src/handlers/api/routes.js`
> 
> ```js
> module.exports = [
>   {
>     method: 'GET',
>     path: '/health',
>     handler: function (request, h) {
>       return h.response({ status: 'OK' }).code(200)
>     },
>     options: {
>       tags
>     }
>   }
> ]
> ```

Hey @lewisdaly the health endpoint in src/handlers/api/routes.js is created for each kafka handler instance, i.e. prepare-handler, fulfil-handler-, position-handler
So I would suggest pointing the handler function in the route to your getHealth function in src/api/root/handler.js

lewisdaly:
> Hey @lewisdaly the health endpoint in src/handlers/api/routes.js is created for each kafka handler instance, i.e. prepare-handler, fulfil-handler-, position-handler
> So I would suggest pointing the handler function in the route to your getHealth function in src/api/root/handler.js

Ah I see. Thanks for the heads up.

lewisdaly:
@mdebarros @rmothilal updated the kafka health checks, please take a look when you get the chance.

src/api/root/handler.js
```diff
@@ -0,0 +1,184 @@
...
+        status: isLocked ? statusEnum.DOWN : statusEnum.OK
+      }
+    }
+    case 'broker': {
+      /*
+        Ensure that all topics for consumers and producers are available
+        Otherwise consider Kafka DOWN
+      */
+      const consumerTopics = [
+        Util.transformGeneralTopicName(TransferEventType.TRANSFER, TransferEventAction.PREPARE),
```
rmothilal: @mdebarros @lewisdaly I believe this will fail when the API is run standalone. This needs to be taken into consideration for the API as well as the API for the Kafka handlers. This will in turn effect the getSubServiceHealth and the evaluateServiceHealth functions. I believe we need to revisit the way this check is done to something more generic

lewisdaly: Sure. I think I need to get a better understanding of running the api in standalone first.

Do we have any tests (integration or otherwise) that test the endpoints in standalone?

rmothilal: Yes @lewisdaly we do have integration tests in test/integration/domain/participant/index.test.js
But i would also look where the initialise function in src/shared/setup.js is called from and the different ways that it is started up. If you need any assistance at any point please feel free to message me or @mdebarros on slack.

lewisdaly:
Closing as this will require rather large changes to the new health check approach. Will open another PR when I'm ready.

PR #303, opened 2019-07-01
Per Sonar recommendations modified code to address 6 bugs and 2 vulnerabilities

gibaros:
2 vulnerabilities were related to console logging 

6 bugs; 5 were related to evaluating element in foreach before executing logic and 1 was related to merging a case in a switch with the same business logic

### DIFF
--- a/src/shared/setup.js
+++ b/src/shared/setup.js
@@ -222,8 +222,6 @@ const initialize = async function ({ service, port, modules = [], runMigrations 
   let server
   switch (service) {
     case 'api':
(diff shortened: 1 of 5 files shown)


src/shared/setup.js
```diff
@@ -222,8 +222,6 @@ const initialize = async function ({ service, port, modules = [], runMigrations
   let server
   switch (service) {
     case 'api':
-      server = await createServer(port, modules)
```
mdebarros: Not sure this should have been removed?

PR #331, opened 2019-07-23
Fix/868 timeout handler error

lewisdaly:
Fixes: https://github.com/mojaloop/project/issues/868

Resolves an issue where the transferId was undefined, causing the positions handler to fail to reset the positions after a timeout.

Also added a new integration tests to test this functionality.


test/integration/handlers/handlers.test.js
```diff
@@ -542,8 +545,68 @@ Test('Handlers test', async handlersTest => {
...
+      test.equal(payerInitialPosition, payerCurrentPosition.value, 'Position resets after a timeout')
+      test.end()
+    })
+
+    timeoutTest.end()
+  })
+
   await handlersTest.test('teardown', async (assert) => {
     try {
+      Handlers.timeouts.stop()
```
ggrg: Handlers.timeouts.stop is async - please add await

src/handlers/positions/handler.js
```diff
@@ -101,7 +101,13 @@ const positions = async (error, messages) => {
     const payload = decodePayload(message.value.content.payload)
     const eventType = message.value.metadata.event.type
     const action = message.value.metadata.event.action
+
     const transferId = payload.transferId || (message.value.content.uriParams && message.value.content.uriParams.id)
+    if (!transferId) {
```
ggrg: Please comment when transferId is undefined. I'm happy to see it is handled, but throwing an error from the cron job might not be the best solution. It should rather not happen at all.

lewisdaly: Yeah good point, I hadn't considered that.

test/integration/handlers/handlers.test.js
lewisdaly: Thanks. Will do

src/handlers/positions/handler.js
ggrg: Thanks for adding entry to the log, but I'm still interested if you have a scenario when this is happening?

lewisdaly: See the attached ticket for this fix: mojaloop/project#868

@mdebarros or @rmothilal might have something to add, but the plan is to change the way these messages are generated to a factory method in the `central-services-shared` or `central-services-stream` library. 

By adding this null check in now, in the future when we switch to the new method, we can be sure that the bug won't come back to bite us.

ggrg: Thanks @lewisdaly, you still haven't replied my question: _Do you have a scenario when this is happening?_ I've seen the attached ticket and without running the code I can see that the following fixes it:
`src/handlers/timeouts/handler.js`
```
          uriParams: {
            id: result[i].transferId
          }
```
And because transferId is not supposed to be undefined under any circumstance for timeouts I'd like to know if 
**A**. you have added this check just in case we experience the same issue in future (e.g. reintroducing the bug with a new feature) or 
**B**. you currently have a test case which you're handling with the check **if (!transferId)**

lewisdaly: **A**

There is no other case where **transferId** is null, but I wanted to make sure we won't have the same problem if/when we move to a factory method for generating the messages.

ggrg: ok, thanks

PR #341, opened 2019-08-07
fix/909-invalid-json

lewisdaly:
- Add --silent flag to npm run audit:check to ensure we print valid JSON
- bump package version to 7.2.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "7.2.0",
+  "version": "7.2.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "7.2.0",
+  "version": "7.2.1",
```
mdebarros: Please, can you bump up to 7.3.0 since we are in that sprint!

mdebarros: I will merge this and make another PR to change this.

PR #348, opened 2019-08-12
Fix for pi7.3 release

mdebarros:
- Updated Participant Update operation to fetch participant data from DB instead of updating the previously returned DB model via a Boolean addition operation

# **DONT MERGE**


mdebarros:
Additional tests have been added to the QA Framework to ensure that there is no inconsistency with regard to the participant isActive status: https://github.com/mojaloop/postman/pull/28

We will only address the changes within this PR if there are any failures from the QA tests.

PR #372, opened 2019-08-29
#954participant_inactive_stop_tranfersTestfailsonSendTransferfail: Hotfix for https://github.com/mojaloop/project/issues/954

mdebarros:
- Removed logTransferError from prepare handler. This should not occur here as the prepare handler would only receive a transfer prepare. Error callbacks are handled by the fulfil handler. Fulfil handler will need to be enhanced to ensure that the Error is logged correctly.


mdebarros:
Replaced with https://github.com/mojaloop/central-ledger/pull/373

PR #380, opened 2019-09-02
temp change logging to verbose

lewisdaly:
- Make this logging verbose temporarily, in order to hopefully catch the issue with ENOAUDIT when running `npm run audit:check`


lewisdaly:
@mdebarros bump

PR #389, opened 2019-09-06
Bugfix/489 Cross Currency Transfer Error

ggrg:
Before:
```
{
  "errorInformation": {
    "errorCode": "2001",
    "errorDescription": "Internal server error - Invalid FSP name or currency",
    "extensionList": {
      "extension": []
    }
  }
}
```
After:
```
{
  "errorInformation": {
    "errorCode": "3100",
    "errorDescription": "Generic validation error - Participant dfsp1 EUR account not found",
    "extensionList": {
      "extension": []
    }
  }
}
```


mdebarros:
@ggrg does this PR resolve --> https://github.com/mojaloop/project/issues/489?

All I see is a single error being commented out.

ggrg:
@mdebarros, yes it does fix it. The explanation is that when transfer has not passed validation, we save as much as we can, without throwing Internal Server Error. Where this explanation was supposed to go?
Please note, this PR is planned to be merged after https://github.com/mojaloop/central-ledger/pull/408.

PR #403, opened 2019-09-17
Release/v8.1.0

ndonnan:
### Updates for release v8.1.0

- Bumped version
- Merged master and resolved conflicts
- Integrated event-sdk and added spans to transfer, position and timeout handlers
- Integrated central-services-logger
- Update routes and tests to support Hapi + Joi 16.x:
  - wrapped payload in Joi.object().
  -  changed .valid() to use expanded parameters instead of an array.


src/handlers/transfers/handler.js
```diff
@@ -86,48 +87,63 @@ const prepare = async (error, messages) => {
...
-    }
     const payload = decodePayload(message.value.content.payload)
     const headers = message.value.content.headers
     const action = message.value.metadata.event.action
     const transferId = payload.transferId
+    span.setTags({ transactionId: transferId })
+    await span.audit(message, EventSdk.AuditEventAction.start)
     const kafkaTopic = message.topic
     let consumer
+    await span.debug({
```
mdebarros: Please comment out in the same approach (with **TODO** comment) as ml-api-adapter, e.g: https://github.com/mojaloop/ml-api-adapter/pull/190/commits/971c1f21e08a72d9a533aeb1d6909ffdcabe6b6e.

Please also do a search for any `span.debug` that need to be similarly updated.

ndonnan: I have commented out as per ml-api-adapter, this was the only span.debug

PR #408, opened 2019-09-21
Feature/749 Bulk Transfers Timeout

ggrg:


ggrg:
> Need to first publish central-services-shared to validate tests.
> 
> I think it would also be wise to wait for the changes that @rmothilal is introducing on the central-services-shared & central-services-stream to move all Kafka specific dependency code out of central-services-shared into central-services-stream.

Thanks @mdebarros - Tests have been validated. I wouldn't like to postpone closing this story and I wouldn't support the idea of stories started later to have priority over those which are already finished.

mdebarros:
> > Need to first publish central-services-shared to validate tests.
> > I think it would also be wise to wait for the changes that @rmothilal is introducing on the central-services-shared & central-services-stream to move all Kafka specific dependency code out of central-services-shared into central-services-stream.
> 
> Thanks @mdebarros - Tests have been validated. I wouldn't like to postpone closing this story and I wouldn't support the idea of stories started later to have priority over those which are already finished.

Is this ready to be reviewed? I see that there have been updates since your comment?

P.s. Vuln check is also failing.

src/handlers/timeouts/handler.js
```diff
@@ -1,44 +1,45 @@
...
- */
+  * @module src/handlers/timeout
+  */
 
 const CronJob = require('cron').CronJob
 const Config = require('../../lib/config')
 const TimeoutService = require('../../domain/timeout')
 const Enum = require('@mojaloop/central-services-shared').Enum
 const Kafka = require('@mojaloop/central-services-shared').Util.Kafka
+const KafkaProducer = require('@mojaloop/central-services-stream').Util.Producer
```
mdebarros: Rename this to Producer

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "8.1.1",
```
mdebarros: Please keep it as 8.1.1 <-- 8.1.0 was the last release

PR #412, opened 2019-09-24
feature/remove-bad-dependencies

lewisdaly:
- remove unused npm-run-all dependency, which removes the spdx-exceptions@2.2.0 package that has a Creative Commons license
- make license-scanner work against production dependencies only
-  move jsDoc to devDependencies, since it is not used in the actual codebase. This will remove the taffydb nested dependency


lewisdaly:
@mdebarros right you are - I searched for `npm-run-all`, but not `run-s` or `run-p`. I'll move it do `devDependencies` however, unless you have any objections?

PR #420, opened 2019-09-26
Hotfix/#700 update for ml api adapter callback notification config changes

oderayi:
Configuration update for `ml-api-adapter` for configuring notification callback to Payee DFSP on transfers finalization.


.ncurc.json
```diff
@@ -1,4 +1,7 @@
 {
   "reject": [
+    "@mojaloop/central-services-health",
```
mdebarros: Why is this needed?

PR #436, opened 2019-10-09
Feature/update to new database library

rmothilal:
Updated central-ledger to use database config object instead of the connection string as previously used


lewisdaly:
Looks good to me, apart from integration and coverage failures.

PR #463, opened 2019-10-28
Update dependencies for central-ledger

jgeewax:
Just the output of running `npx ncu -u`.


yosheeck:
I guess this is quite old and all proposed bumps/updates are already there.
Thank you for this PR, sorry it wasn't handled on time :)

PR #500, opened 2019-12-05
Feature/#37implement authorization endpoints

DonChangfoot:
Adds the FSPIOP_CALLBACK_URL_AUTHORIZATIONS and FSPIOP_CALLBACK_URL_TRANSACTION to the endpointType seeds. Also updated the populate test data test script to include these endpoint types.


seeds/endpointType.js
```diff
@@ -129,6 +129,14 @@ const endpointTypes = [
...
     name: 'FSPIOP_CALLBACK_URL_BULK_TRANSFER_ERROR',
     description: 'Participant callback URL to which bulk transfer error notifications can be sent'
+  },
+  {
+    name: 'FSPIOP_CALLBACK_URL_AUTHORIZATIONS',
+    description: 'Participant callback URL to which authorization requests can be sent'
+  },
+  {
+    name: 'FSPIOP_CALLBACK_URL_TRANSACTION',
+    description: 'Participant callback URL to which transaction requests can be sent'
```
elnyry: @DonChangfoot we have a /transactions resource so may be this will be misleading.. how about something else such as FSPIOP_CALLBACK_URL_TRX_REQ_SERVICE?

elnyry: this will affect the tests as well..

DonChangfoot: @elnyry ok, the `transaction-request-service` currently uses the `FSPIOP_CALLBACK_URL_TRANSACTION` for the transaction requests. Should that be the case or shall I update it to use `FSPIOP_CALLBACK_URL_TRX_REQ_SERVICE`?

elnyry: @DonChangfoot I think FSPIOP_CALLBACK_URL_TRX_REQ_SERVICE makes sense to avoid ambiguity

DonChangfoot: @elnyry Can you please look at https://github.com/mojaloop/central-services-shared/pull/159 then I can make the changes for this repo.

elnyry: @DonChangfoot PR-159 approved..

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "8.8.1",
+  "version": "8.8.2",
```
elnyry: 8.8.1 is still fine, we don't have a release for that yet

DonChangfoot: ok cool. I've changed it back.

PR #543, opened 2020-01-10
#1163 - Allow simple health check

oderayi:
- Enable switching between simple and detailed health checks
- Updated dependencies

### DIFF
--- a/src/api/root/handler.js
+++ b/src/api/root/handler.js
@@ -26,17 +26,7 @@
 
 const HealthCheck = require('@mojaloop/central-services-shared').HealthCheck.HealthCheck
 const { defaultHealthHandler } = require('@mojaloop/central-services-health')
-
 const packageJson = require('../../../package.json')
-const {
-  getSubServiceHealthDatastore,
-  getSubServiceHealthBroker
-} = require('../../lib/healthCheck/subServiceHealth')
-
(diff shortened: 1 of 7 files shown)


src/api/root/handler.js
```diff
@@ -46,7 +36,24 @@ const healthCheck = new HealthCheck(packageJson, [
  * @param {*} request - the Hapi request object
  * @param {*} h - the Hapi handler object
  */
-const getHealth = defaultHealthHandler(healthCheck)
+const getHealth = (request, h) => {
+  const simpleHealthCheck = request.query && 'simple' in request.query && (request.query.simple === '' || request.query.simple === true)
+  let healthCheck
+  if (simpleHealthCheck) {
+    healthCheck = new HealthCheck(packageJson, [])
```
lewisdaly: Ironically, creating a new HealthCheck object on every `getHealth` request is much less efficient than just recycling the existing one like the old implementation.

I would suggest initializing 2 HealthCheck objects: `healthCheck` and `simpleCheck` in the global scope of this file, and then:
```js
if (simpleHealthCheck) {
  return defaultHealthHandler(simpleCheck)(request, h)
}

   return defaultHealthHandler(healthCheck)(request, h) 
```

PR #567, opened 2020-01-23
Feature/remove spdx exceptions package

lewisdaly:
This is a somewhat hacky fix, but it does remove a deeply nested package that we don't use which is incompatible with our license policy.

Here's the `npm ls` of the `spdx-exceptions` package:
```
@mojaloop/central-ledger@8.8.2 /Users/lewisdaly/developer/vessels/mojaloop/central-ledger
└─┬ npm-run-all@4.1.5
  └─┬ read-pkg@3.0.0
    └─┬ normalize-package-data@2.5.0
      └─┬ validate-npm-package-license@3.0.4
        └─┬ spdx-expression-parse@3.0.0
          └── spdx-exceptions@2.2.0
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "8.8.2",
+  "version": "8.8.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


lewisdaly:
I was wrong, turns out `npm-run-all` does need this package...

PR #576, opened 2020-01-29
#602: refactored participant Cache

yosheeck:
Refactor of existing cache implementation after discussion with @mdebarros and @rmothilal.
The context is that lib/cache is now getting correctly "libraried", so it will not depend on anything that's being cached. Cache is now becoming "generic" lib where all what's cached has to register as cache-client, with dependency injection of things like data preload or flush actions.
As for this PR, the is true for 'participants' cache, but still TODO for enums (planned to be done in incoming PRs).


src/models/participant/participantCached.js
```diff
@@ -0,0 +1,141 @@
...
+    return cachedParticipants.allParticipants
+  } catch (err) {
+    throw ErrorHandler.Factory.reformatFSPIOPError(err)
+  }
+}
+
+const withInvalidate = (theFunctionName) => {
+  return async (...args) => {
+    try {
+      const result = await ParticipantModel[theFunctionName](...args)
```
mdebarros: Should you not use the cache mechanism to retrieve the data once the cache has been invalidated?

PR #607, opened 2020-02-20
feature/1218 Settlement Model Handler Design

ggrg:
- Added transferParticipantStateChange table and related FKs and indexes

_**Note**: Please, ignore the long lists of commits below. It's due to recent merge of settlement-v2 and re-basing this PR to master_


ggrg:
I have pushed the above as a [branch](https://github.com/mojaloop/central-ledger/tree/feature/1218SettlementModelHandler) to mojaloop/central-ledger, so that the team will be able to manage it in my absence after the DA review.

PR #615, opened 2020-02-26
Settlement-v2 merge into master

ggrg:
Probably, it is good time to consider merging into master to avoid further maintenance


migrations/910102_hotfix954.js
```diff
@@ -25,33 +25,7 @@
...
- - Name Surname <name.surname@gatesfoundation.com>
-
- * Miguel de Barros <miguel.debarros@modusbox.com>
- --------------
- ******/
-
-'use strict'
-
-exports.up = async (knex) => {
+exports.up = async (knex, Promise) => {
```
rmothilal: Why did you remove the license text?

seeds/z1000_settlementModel.js-
```diff
@@ -0,0 +1,56 @@
+/*****
```
rmothilal: why does the filename have a dash at the end of it?

ggrg: It is currently not executed, but I just wanted to preserve it for a while until it's absolutely certain, there won't be any generic settlement models to be seeded. Please, let me know if it would be better to delete the script.

rmothilal: I would make a note in the file if you wish to keep it

migrations/910102_hotfix954.js
rmothilal: aha thanks

seeds/z1000_settlementModel.js-
rmothilal: I would suggest making a Note in the file stating why it is there and that it is not used. also consider appending or prepending the file name with ignore/deprecated or an appropriate term

PR #616, opened 2020-02-26
Fix/1107 circleci deploy

lewisdaly:
- Update CircleCI deploy config to deploy the quoting-service after a release
- uses new circleci orb to accomplish this
- remove redundant duplicated CircleCI config
- currently deploys to testci namespace on dev1 cluster, but this can be modified from CircleCI params
- part of #1107


.circleci/config.yml
```diff
@@ -383,43 +292,55 @@ jobs:
...
-          <<: *defaults_deploy_config_kubernetes_context
-      - run:
-          <<: *defaults_deploy_set_kubernetes_context
-      - run:
-          <<: *defaults_deploy_configure_helm
-      - run:
-          <<: *defaults_deploy_install_or_upgrade_helm_chart
+      - deploy-kube/setup_and_run:
+          helm_set_values: |
+            --set .central.centralledger.centralledger-service.containers.api.image.repository=$DOCKER_ORG/$CIRCLE_PROJECT_REPONAME \
```
lewisdaly: these need to not start with a `.`

lewisdaly: these need to not start with a `.`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "9.2.1",
+  "version": "9.2.2",
```
rmothilal: this should be 9.2.3

PR #647, opened 2020-03-25
updated central-services-shared and event-sdk versions to support cus…

vgenev:
…tom tags in tracestate


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "9.4.0",
+  "version": "9.4.1",
```
mdebarros: Please revert to 9.4.0 as we have not made a release as of yet.

vgenev: ok

Issue #174, opened 2018-10-15
Some postman tests do not work


GeekGuste:
Hello,
I followed the procedure of installation of the central ledger until the start of the server. But the test with the file Mojaloop.postman_collection.json was inconclusive. The creation of the participants works but the Set Limits and Initial Position for PayerFSP and other tests do not work. After going through the project code a little bit, I noticed that the test paths that returned errors were not reported. So I think the project is not up to date. I would like you to reassure me if I am wrong or if not, please review this part of the project.
Regards

elnyry:
Hi _GeekGuste_

Please let me know which resources/docs you used to run central ledger locally. It was likely incomplete.

Also, the title is misleading, the Postman tests **do** work - with the right configuration.

GeekGuste:
Hi elnyry

Thank you for your answer.

I used this document [https://github.com/mojaloop/central-ledger/blob/master/Onboarding.md](https://github.com/mojaloop/central-ledger/blob/master/Onboarding.md) to run central ledger and I am using the file [https://github.com/mojaloop/postman/blob/master/Mojaloop.postman_collection.json](https://github.com/mojaloop/postman/blob/master/Mojaloop.postman_collection.json) for tests. Also I did not install the central ledger locally. I did it on a vps. And some postman tests do not work, for exemple the creation of the participants works but the Set Limits and Initial Position for PayerFSP is returning a NotFoundError.

elnyry:
hi,

Thanks for the details. The admin/operations API is a moving target right now as a lot of work is being done on it to provide Settlements/Accounts features. I understand that there were probably changes to the API and the postman collection was probably not updated.

However, we are currently going to streamline this process to make sure that both of these are in sync. Will get back to you soon (hopefully, in a week), once this is fixed.

Thank you,
Sam

elnyry:
We now have a postman repo that has decent coverage: https://github.com/mojaloop/postman

sri-miriyala:
@elnyry  With the latest version 5.1.0 , this collection needs to be updated especially for admin endpoints. Once I complete validating, I will update this collection.

elnyry:
Hi I think now the collections, tests in 'postman' repo are in decent shape, so closing this. Please open a bug if there is a specific issue.

GeekGuste:
Ok

PR #651, opened 2020-03-26
quotes-service issue #174

bushjames:
quotes-service issue #174 make quoteResponseId column of quoteExtension table nullable.


lewisdaly:
This looks pretty outdated to me. Please update and open a new PR if it's still relevant. We also might need to remove Rajiv from the git commit history.

PR #659, opened 2020-04-06
Feature/#1263 transfer metrics

vgenev:
* Merged the metrics from https://github.com/mojaloop/central-ledger/tree/release/v9.5.0-perf1.0
* Added index to optimise transferParticipant lookup's as part of the Position Management during the Transfer process


docker-compose.yml
```diff
@@ -150,16 +150,18 @@ services:
       start_period: 40s
 
   simulator:
-    image: mojaloop/simulator:latest
+    image: mojaloop/simulator-perf:m9.2.1-perf4.1.2
```
mdebarros, line 153: Please use an official Simulator release. I.e. if we need, we will have to also migrate perf changes into master branch.

migrations/310203_transferParticipant-indexes.js
```diff
@@ -0,0 +1,38 @@
...
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ * ModusBox
+ - Georgi Georgiev <georgi.georgiev@modusbox.com>
+ --------------
+ ******/
+
+'use strict'
+
+exports.up = function (knex) {
```
mdebarros, line 28: Make sure you include a note as to why this was added as part of the PR.

I.e. Added index to optimise transferParticipant lookup's as part of the Position Management during the Transfer process

PR #681, opened 2020-04-15
Monitor the mongodb port in docker-compose instead since mongodb has …

ndonnan:
…been removed from alpine packages


docker/central-ledger/default.json
```diff
@@ -3,7 +3,7 @@
   "HOSTNAME": "http://central-ledger",
   "DATABASE": {
     "DIALECT": "mysql",
-    "HOST": "localhost",
+    "HOST": "mysql",
```
rmothilal, line 6: this correct?

PR #686, opened 2020-04-21
Fix/test util scripts

georgi-logodazhki:


test/util/scripts/env.sh
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
-export KAFKA_ID=kafka
+export KAFKA_ID=cl_kafka
```
rmothilal, line 2: kafka is not specific for central-ledger ml-api-adapter also uses the same instance

test/util/scripts/env.sh
```diff
@@ -1,13 +1,13 @@
...
 export MOCKSERVER_ID=mockserver
 export MOCK_SIM_PORT=8444
-export DB_ID=mysql
+export DB_ID=cl_mysql
 export DBUSER=central_ledger
 export DBPASS=password
 export DBNAME=central_ledger
 export DB_SLEEPTIME=15
-export OBJ_ID=objstore
+export OBJ_ID=cl_objstore
```
rmothilal, line 10: mongodb is not specific for central-ledger ml-api-adapter also uses the same instance

PR #708, opened 2020-05-13
Fix/ml_user can't invoke apk in wait-for

eoln:
So I go back to `root` user.
I tried adduser with --system flag - but it doesn't work


lewisdaly:
3 thoughts:

1. We should rewrite the wait for scripts to be javascript that uses existing `node_modules` instead of relying 
2. Running our docker users as `root` is a security risk for the host system, and not recommended (hence why we created the new user)
3. Did you try running the docker container as root? You can do this in docker-compose with `user: root` underneath the service. I found that it works for me.

eoln:
> 3 thoughts:
> 
> 1. We should rewrite the wait for scripts to be javascript that uses existing `node_modules` instead of relying

Yes, you are right, maybe there is a need to rewrite these scripts or simply add mysql-client to the container's image. I think it shouldn't increase a lot of the size of the container's image.

> 2. Running our docker users as `root` is a security risk for the host system, and not recommended (hence why we created the new user)

...aha...

> 3. Did you try running the docker container as root? You can do this in docker-compose with `user: root` underneath the service. I found that it works for me.

Yes I tried `user:root` and it works well for me too :-D  
But I lost an hour to figure out why it isn't working for the version we have here: https://github.com/mojaloop/pisp/blob/master/docker-local/docker-compose.yml#L9 So, I created this PR to keep docker-compose scripts working.

eoln:
> So I go back to `root` user.
> I tried adduser with --system flag - but it doesn't work

After a small talk with @lewisdaly I changed approach to keep `ml-user` instead of `root`. 
So I propose a new way of running `wait-for` scripts for docker-compose approach.

Dockerfile
```diff
@@ -30,5 +30,7 @@ USER ml-user
 COPY --chown=ml-user --from=builder /opt/central-ledger .
 RUN npm prune --production
 
+RUN npm prune --production
```
lewisdaly, line 33: We don't need this twice.

eoln: done

lewisdaly:
@eoln is this outdated now? Can it be closed?

PR #721, opened 2020-05-21
add migration to make quoteResponseId nullable in quoteExtension table

bushjames:
Quote service writes extension lists on quote requests and responses to quoteExtension table. For quote requests, the quoteResponseId is not known so this column needs to be nullable.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "10.1.1",
+  "version": "10.1.2",
```
rmothilal, line 3: this would be 10.2.0

rmothilal: also please run npm install after that is changed

PR #732, opened 2020-06-03
Feature/OTC-525 Implement GET transaction Object by transferId

lazolalucas:


src/domain/transactions/index.js
```diff
@@ -0,0 +1,59 @@
...
+  const jsonPacket = await decodeIlpPacket(inputIlpPacket)
+  return JSON.parse(jsonPacket)
+}
+/**
+ * Decodes an Ilp Packet
+ *
+ * @returns {object} - Ilp packet as JSON object
+ */
+const decodeIlpPacket = async function (inputIlpPacket) {
+  const binaryPacket = Buffer.from(inputIlpPacket, 'base64')
```
ndonnan, line 53: I'm not sure the packet as defined in the spec can be decoded in this way, check how it is done in the SDK Standard Components here: https://github.com/mojaloop/sdk-standard-components/blob/master/src/lib/ilp/index.js#L214

PR #734, opened 2020-06-08
Edited CI to build PISP docker image.

kleyow:


kleyow:
Addressed versioning concerns.

PR #751, opened 2020-06-29
Feature/#1334 patch request notif

vgenev:


.env
```diff
@@ -15,3 +15,4 @@ EVENT_SDK_VENDOR_PREFIX="mojaloop"
 EVENT_SDK_TRACESTATE_HEADER_ENABLED="false"
 EVENT_SDK_ASYNC_OVERRIDE_EVENTS="log, trace"
 EVENT_SDK_TRACEID_PER_VENDOR="false"
+CLEDG_HANDLERS_TIMEOUT_DISABLED="true"
```
rmothilal, line 18: is this supposed to be here? that config is in default.json

PR #753, opened 2020-07-03
test: poc to prove unhandled rejection is bubbling out from unit test

eoln:
I picked random test to make this POC.

Below is a proof that unhandled promise rejection is bubbled out from unit test

I also fixed the test so now it fails.

If you comment out the newly added L197 and run the test in console output you will see the message from `process.on` handler which will be called if the unit test code is not handling the error properly.


lewisdaly:
I'm going to close this one. It still acts as a good reference to the mentioned issue.

PR #757, opened 2020-07-09
OTC-643 adding settlement continual delay type

claudio-viola:


claudio-viola:
closing existing enum values suffice to cover use cases

PR #764, opened 2020-07-22
OTC-728 adding ledgerAccountTypes api create and get

claudio-viola:
- added GET /ledgerAccountTypes to get all ledger account type
- added POST /ledgerAccountTypes to create a new ledger account type


claudio-viola:
closing for the time being.

PR #770, opened 2020-08-11
#1547: Ignore "RESERVED" transferState from v1.0 clients on fulfill callback

oderayi:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "11.0.2",
+  "version": "11.0.3",
```
elnyry-sam-k, line 3:
```suggestion
  "version": "11.1.0",
```

PR #779, opened 2020-08-31
chore: add thirdparty endpoints to database seeds

kleyow:


migrations/20200831143314_endpointNameLength.js
```diff
@@ -0,0 +1,46 @@
+/*****
```
kleyow, line 1: unsure what file naming scheme the migration uses :thinking: `migrate:create` give me `20200831143314...js where as other migrations are some different numbering scheme.

migrations/20200831143314_endpointNameLength.js
```diff
@@ -0,0 +1,46 @@
...
+ --------------
+ ******/
+
+'use strict'
+
+exports.up = async (knex) => {
+  return await knex.schema.hasTable('endpointType').then(function(exists) {
+    if (exists) {
+      return knex.schema.alterTable('endpointType', function(t) {
+        t.string('name', 100).alter();
```
kleyow, line 32: Is it fine for us to make this `VARCHAR` bigger or do we need to make our endpoint names smaller > - >

eoln: maybe let keep it as is and shorten THIRDPARTY prefix to TP... ?

eoln: and CALLBACK to CB

kleyow: @lewisdaly @sridharvoruganti  any opinion on this? length migration vs endpoint refactor?

sridharvoruganti: either is fine for me, but if we rename these endpoints then code changes required in central-services-shared and thirdparty-api-adapter.

lewisdaly: Yeah let's shorten the names here - migrations in central-ledger should be avoided if at all possible

kleyow: done. shortened the endpoint names Thirdparty -> TP. Callback -> CB. Authorizations -> Auth.

I'll make the changes in central-services-shared and thirdparty-api-adapter after this is merged.

eoln (approved):
lgtm

PR #781, opened 2020-09-03
refactor: update pisp/master

kleyow:


lewisdaly (approved):
Thanks Kevin!

kleyow:
Awesome. I was able to pull this branches and confirm that this fixes the mimetype error I was encountering.
That'll help me wrap up my current work. Still planning on upgrading docker-local services lest we drag too far behind master.

PR #792, opened 2020-11-16
BugFix 1444 - Ignore Dups on Seed insert

shashi165:
### Summary:
While inserting the new seeds, if there are any existing values in the table then it was failing silently without inserting the new seed values.
Changed the insert query to use `INSERT IGNORE` to ignore the dups and continue adding the new values. Also updated the unit tests.

### Unit Tests:
```
1..1378
# tests 1378
# pass  1378

# ok
```
### Coverage Tests
```
=============================== Coverage summary ===============================
Statements   : 98.94% ( 3273/3308 )
Branches     : 97.65% ( 1204/1233 )
Functions    : 99.5% ( 395/397 )
Lines        : 98.94% ( 3254/3289 )
================================================================================
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "pre-commit": "1.2.2",
     "proxyquire": "2.1.3",
     "sinon": "9.1.0",
-    "standard": "16.0.2",
+    "standard": "14.3.4",
     "tap-xunit": "2.4.1",
     "tape": "4.13.2",
     "tapes": "4.1.0"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "11.2.4",
+  "version": "11.3.0",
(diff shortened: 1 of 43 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "11.2.3",
+  "version": "11.2.4",
```
elnyry-sam-k, line 3: I think we can consider this new functionality.. 11.3.0?

vgenev (dismissed):

package.json
vgenev: I am ok after that is addressed..

mdebarros (dismissed):
+1

package.json
shashi165: Done

vgenev (approved):
👍

elnyry-sam-k (dismissed):

vgenev (approved):
go!

PR #817, opened 2021-02-21
feat: merge pisp thirdparty endpoint seeds from pisp/master

kleyow:


package.json
```diff
@@ -15,7 +15,9 @@
     "Sam Kummary <sam@modusbox.com>",
     "Steven Oderayi <steven.oderayi@modusbox.com>",
     "Valentin Genev <valentin.genev@modusbox.com>",
-    "Shashikant Hirugade <shashikant.hirugade@modusbox.com>"
+    "Shashikant Hirugade <shashikant.hirugade@modusbox.com>",
+    "Paweł Marzec <pawel.marzec@modusbox.com>",
+    "Kevin Leyow <kevin.leyow@modusbox.com>"
```
eoln, line 20: what about bumping the version?

eoln (changes requested):
lgtm, the package version needs to be bumped manually...

package.json
kleyow: I was hoping to have https://github.com/mojaloop/central-ledger/pull/818 to do it for me but now that I think of it, it wouldn't catch this `feat` in the bump.

done and ty.

eoln (approved):
lgtm

PR #823, opened 2021-02-28
docs: add overview of automated releases in readme

kleyow:


README.md
```diff
@@ -111,3 +112,32 @@ For more information on anchore and anchore-cli, refer to:
...
+push a release triggering another subsequent build that also publishes a docker image.
+
+### Potential problems
+
+*   There is a case where the merge to master workflow will resolve successfully, triggering
+    a release. Then that tagged release workflow subsequently failing due to the image scan,
+    audit check, vulnerability check or other "live" checks.
+
+    This will leave master without an associated published build. Fixes that require
+    a new merge will essentially cause a skip in version number or require a clean up
```
mdebarros, line 134: Would it not be possible to check if there is an associated release?

I.e. only increment the version if there was a successful publish.

README.md
```diff
@@ -111,3 +112,32 @@ For more information on anchore and anchore-cli, refer to:
...
+    of the master branch to the commit before the CHANGELOG and bump.
+
+    This may be resolved by relying solely on the previous checks of the
+    merge to master workflow to assume that our tagged release is of sound quality.
+    We are still mulling over this solution since catching bugs/vulnerabilities/etc earlier
+    is a boon.
+
+*   It is unknown if a race condition might occur with multiple merges with master in
+    quick succession, but this is a suspected edge case.
```
mdebarros, line 143: This could only be an issue if a later build finishes prior to an earlier build, or can you think of any other situation where this would be possible?

I do not believe CircleCI has the ability to ensure the above consistency, in fact here is a feature request for that very purpose: https://discuss.circleci.com/t/prevent-concurrent-builds-of-a-branch/1142

But as you can see from that request, CircleCI does cancel redundant builds if configured as such. We currently have this enabled. This should help ensure that this does not occur.

mdebarros: https://github.com/bellkev/circle-lock-test <-- there is an apparent work-around.

millerabel:
If I reason about this as a release candidate that is pulled and built (after a push to the main branch), it’s not yet tagged. So we must have built the untagged images, yes? The images should be scanned and post build steps completed before the release tag is pushed. 

I think we’re doing all that. 


If you get this far, the final step is to be sure we tag the release and that the tagged release can be pulled and built. 


If building the tagged release fails to verify, someone has to remedy the failure manually, push a patch, then re-verify (by pulling the patched release candidate and rebuilding, scanning, and testing the images).

The final release tag would then read something like “17.3.1” rather than “17.3.0”. This image is released as “17.3” for external communications, but it is released with a patch, and so the patch level for the 17.3 release is already at “.1” Others will surely follow. :-)

So the question, if I pull 17.3, do I get the latest patch level of each file (17.3.1 in this hypothetical)?

And do we replace previous patch level verified images with only last known good? (If not, why not?)

So if someone uses the last known good docker images, there is a patch-level matching tagged release, even as work proceeds on the main branch. 

I realize I may not fully understand the issue or the disconnect, please point out my errors!

-- Miller


On Mar 5, 2021, at 5:08 AM, Miguel de Barros <notifications@github.com> wrote:

﻿
@mdebarros commented on this pull request.

In README.md:

> +and image builds. This process essentially mimics a manual tag and release.
+
+On a merge to master, CircleCI is configured to use the mojaloopci github account
+to push the latest generated CHANGELOG and package version number.
+
+Once those changes are pushed, CircleCI will pull the updated master, tag and
+push a release triggering another subsequent build that also publishes a docker image.
+
+### Potential problems
+
+*   There is a case where the merge to master workflow will resolve successfully, triggering
+    a release. Then that tagged release workflow subsequently failing due to the image scan,
+    audit check, vulnerability check or other "live" checks.
+
+    This will leave master without an associated published build. Fixes that require
+    a new merge will essentially cause a skip in version number or require a clean up
Would it not be possible to check if there is an associated release?

I.e. only increment the version if there was a successful publish.

kleyow:
| So the question, if I pull 17.3, do I get the latest patch level of each file (17.3.1 in this hypothetical)?

Yes.

| “17.3” for external communications

I wasn't sure how external parties handle their versioning, but if they target versions like that they shouldn't have to worry as long as the latest patch has been built properly

| And do we replace previous patch level verified images with only last known good? (If not, why not?)

Not entirely sure what is being said here. If say 17.3.1 failed to publish an artifact and 17.3.2 published properly. 17.3.1 would have no associated artifact and would need manual intervention if we want to make sure we always have a 1:1 relationship.

| So if someone uses the last known good docker images, there is a patch-level matching tagged release, even as work proceeds on the main branch.

If I'm understanding your statement correctly then yes.

Hmm. Maybe some build infographics would help us...

kleyow:
Forgive me at this awful attempt to visualize what I'm thinking are edgecases.

release_flow

elnyry-sam-k:
@kleyow - would it be possible to bring this up to date, so that we can merge it? If not its getting stale and we may have to close it

kleyow:
@elnyry-sam-k I think the commits applied still apply to automated releases but I don't think I really addressed comments. At the end of the day the the CI really just automates how we currently do releases but maybe not in the same order...in retrospect I should try and change it.

Only possible way to update this to address comments is to improve and give the CI another pass to eliminate edge cases.
The comments still stand but haven't really happened in the year we've had it churning some of our builds.

Manually
- Push change with version bump included. 
- CI PR checks pass. 
- Merge into master. 
- Master CI checks pass. 
- Tag version. 
- Push version to Github. 
- Version CI pass. 
- Release made in github releases.

Automated
- Push change no version bump.  
- CI PR checks pass. 
- Merge into master. 
- Master CI checks pass and CI pushes new commit bumping package version/changes based on commit. 
- New commit skipped since it's just a non functional bump. 
- CI tags and pushes tag. 
- Standardized release and notes made in github releases before Version CI passes (this should be re-ordered). 
- Version CI pass. All is good.

kleyow:
Feel free to merge it, close it, make a story for improvements or story to remove the automated ci.

PR #824, opened 2021-03-01
fix: #1977 timeout enumeration for cron job fixed

vgenev:
Fixed enumeration error for cron job timeout value


src/lib/config.js
```diff
@@ -17,7 +17,6 @@ module.exports = {
   HANDLERS_CRON: RC.HANDLERS.CRON,
   HANDLERS_CRON_DISABLED: RC.HANDLERS.CRON.DISABLED,
   HANDLERS_CRON_TIMEXP: RC.HANDLERS.CRON.TIMEXP,
-  HANDLERS_CRON_TIMEZONE: RC.HANDLERS.CRON.TIMEZONE,
```
elnyry-sam-k, line 20: Its still there in default.json file
```
    "CRON": {
      "DISABLED": false,
      "TIMEXP": "*/10 * * * * *",
      "TIMEZONE": "UTC"
    },
```

In addition, would like to understand both CRON and TIMEOUT values, specifically what the TIMEXP and TIMEZONE on both mean (if there's any duplication)

mdebarros: @vgenev I dont think we use:
```
  HANDLERS_CRON: RC.HANDLERS.CRON,
  HANDLERS_CRON_DISABLED: RC.HANDLERS.CRON.DISABLED,
  HANDLERS_CRON_TIMEXP: RC.HANDLERS.CRON.TIMEXP,
```

Thus that and the config from default.json should be removed:

```
  "HANDLERS": {
   ....,
    "CRON": {
      "DISABLED": false,
      "TIMEXP": "*/10 * * * * *",
      "TIMEZONE": "UTC"
    },
```

Please confirm if this is correct, and if so, then remove.

mdebarros (dismissed):
Looks good!

But I think this is an opportunity to clean up the config a bit.

Please check if the `HANDLERS_CRON_*` config is still used and if not then please remove it.

elnyry-sam-k (approved):

src/lib/config.js
vgenev: removed

mdebarros:
> Looks good!
> 
> But I think this is an opportunity to clean up the config a bit.
> 
> Please check if the `HANDLERS_CRON_*` config is still used and if not then please remove it.

@vgenev do you not need to remove the config from the integration docker configs as well?

mdebarros (approved):
+1

PR #827, opened 2021-03-11
Add timezone to database connection options set to UTC

noopurj:
mojaloop/project#892


elnyry-sam-k:
vuln check is failing, even after all the interventions..

please use '`npm audit fix` and/or `resolve-audit` to fix the issues..

elnyry-sam-k:
@noopurj - Thanks for the PR..

Would it be possible to bring this up to date, so that we can merge it? If not its getting stale and we may have to close it

PR #835, opened 2021-04-26
feat: #2123 Default settlement model

vgenev:
Added support for default settlement model


package.json
```diff
@@ -91,7 +91,11 @@
     "@mojaloop/central-services-health": "11.0.0",
     "@mojaloop/central-services-logger": "10.6.0",
     "@mojaloop/central-services-metrics": "9.5.0",
+<<<<<<< HEAD
```
lewisdaly, line 94: merge conflicts

seeds/ledgerAccountType copy.js
```diff
@@ -0,0 +1,75 @@
+/*****
```
lewisdaly, line 1: Please fix this filename or remove - I'm not sure if this was accidentally copied.

src/lib/configDataSeeder.js
```diff
@@ -0,0 +1,134 @@
...
+ optionally within square brackets <email>.
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ * Claudio Viola <claudio.viola@modusbox.com>
+ --------------
+******/
+'use strict'
+
+// const LedgerAccountTypeModel = require('../models/ledgerAccountType/ledgerAccountType')
```
lewisdaly, line 26: nit: remove commented lines

src/lib/configDataSeeder.js
```diff
@@ -0,0 +1,134 @@
...
+const ErrorHandler = require('@mojaloop/central-services-error-handling')
+const Config = require('./config')
+const SettlementService = require('../domain/settlement')
+const SettlementModel = require('../models/settlement/settlementModel')
+
+// const LedgerAccountTypesService = require('../domain/ledgerAccountTypes')
```
lewisdaly, line 32: nit: remove commented lines

src/lib/configDataSeeder.js
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/**
+ * [initializeSeedData Adds configurable seeds data]
+ */
+async function initializeSeedData () {
+  const knex = Db.getKnex()
+  return knex.transaction(async trx => {
+    try {
+      // await initializeLedgerAccountTypes(trx)
```
lewisdaly, line 75: nit: remove commented lines

src/lib/configDataSeeder.js
```diff
@@ -0,0 +1,134 @@
...
+  })
+}
+
+async function initializeSettlementModels (trx) {
+  const settlementModelNamesToCreate = Config.SETTLEMENT_MODELS.map(item => {
+    return SETTLEMENT_MODELS_ALIASES[item].name
+  })
+  let existingSettlementModelsNames = await SettlementModel.getSettlementModelsByName(settlementModelNamesToCreate, trx)
+  existingSettlementModelsNames = existingSettlementModelsNames.map(record => record.name)
+  const missingAccountTypes = Config.SETTLEMENT_MODELS.filter(item => {
```
lewisdaly, line 92: can use inline syntax instead of wrapping in `{ }`

src/lib/configDataSeeder.js
```diff
@@ -0,0 +1,134 @@
...
+    const recordsToCreate = missingAccountTypes.map(item => {
+      return SETTLEMENT_MODELS_ALIASES[item]
+    })
+    await Promise.all(recordsToCreate.map(async record => {
+      return SettlementService.createSettlementModel(record, trx)
+    }))
+  }
+}
+
+// async function initializeLedgerAccountTypes (trx) {
```
lewisdaly, line 105: nit: remove commented lines

src/lib/configDataSeeder.js
```diff
@@ -0,0 +1,134 @@
...
+      await trx.commit
+    } catch (err) {
+      await trx.rollback
+      throw ErrorHandler.Factory.reformatFSPIOPError(err)
+    }
+  })
+}
+
+async function initializeSettlementModels (trx) {
+  const settlementModelNamesToCreate = Config.SETTLEMENT_MODELS.map(item => {
```
lewisdaly, line 87: What if `Config.SETTLEMENT_MODELS` isn't set? Config should be backwards compatible. Maybe we can have some sensible defaults, or _explicitly_ make a breaking change.

src/lib/configDataSeeder.js
lewisdaly, line 92: What if `Config.SETTLEMENT_MODELS` isn't set? Config should be backwards compatible. Maybe we can have some sensible defaults, or _explicitly_ make a breaking change.

src/lib/configDataSeeder.js
```diff
@@ -0,0 +1,134 @@
...
+
+      await trx.commit
+    } catch (err) {
+      await trx.rollback
+      throw ErrorHandler.Factory.reformatFSPIOPError(err)
+    }
+  })
+}
+
+async function initializeSettlementModels (trx) {
```
lewisdaly, line 86: This type of function feels more like it belongs in config instead of here... just my $0.02

lewisdaly (changes requested):

src/lib/configDataSeeder.js
vgenev: it's kinda config.. data.. seeder :D
I am not sure where would suit it better...

src/models/position/facade.js
```diff
@@ -52,7 +52,12 @@ const prepareChangeParticipantPositionTransaction = async (transferList) => {
     const knex = await Db.getKnex()
     const participantName = transferList[0].value.content.payload.payerFsp
     const currencyId = transferList[0].value.content.payload.amount.currency
-    const settlementModel = await SettlementModelCached.getByLedgerAccountTypeId(Enum.Accounts.LedgerAccountType.POSITION)
+    const allSettlementModels = await SettlementModelCached.getAll()
+    let settlementModels = allSettlementModels.filter(model => model.currencyId === currencyId)
+    if (!(Array.isArray(settlementModels)) || settlementModels.length <= 0) {
+      settlementModels = allSettlementModels.filter(model => model.currencyId === null) // Default settlement model
+    }
```
mdebarros, line 59: What happens if we don't find a default settlementModel?

mdebarros (changes requested):
Looks good, some minor comments for your attention.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "13.2.6",
+  "version": "13.3.0",
```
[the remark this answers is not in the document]
vgenev: Well only the currency being not obligatory is a change, but won't break a working solution, so I think that the minor is what we need to change in the version.. semver was like: major.minor.patch - major for breaking, minor for nonbreaking functional changes and patch for bugfixes, as I remember..

src/api/settlementModels/routes.js
```diff
@@ -71,7 +71,7 @@ module.exports = [
           settlementGranularity: Joi.string().required().valid(...settlementGranularityList).description('Granularity type for the settlement model GROSS or NET'),
           settlementInterchange: Joi.string().required().valid(...settlementInterchangeList).description('Interchange type for the settlement model BILATERAL or MULTILATERAL'),
           settlementDelay: Joi.string().required().valid(...settlementDelayList).description('Delay type for the settlement model IMMEDIATE or DEFERRED'),
-          currency: Joi.string().required().valid(...currencyList).description('Currency code'),
+          currency: Joi.string().valid(...currencyList).description('Currency code'),
```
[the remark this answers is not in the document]
vgenev: Will do it now..

src/domain/settlement/index.js
```diff
@@ -47,10 +44,13 @@ const createSettlementModel = async (settlementModel, trx = null) => {
...
 
-    // create the accounts required for the settlementModel for existing participants
-    await ParticipantService.createAssociatedParticipantAccounts(settlementModel.currency, ledgerAccountType.ledgerAccountTypeId, trx)
-    await ParticipantService.createAssociatedParticipantAccounts(settlementModel.currency, settlementAccountType.ledgerAccountTypeId, trx)
-
+    if (settlementModel.currency) {
+      // check for existing hub account with the settlementModel to be able to create participant accounts automatically
+      await ParticipantService.validateHubAccounts(settlementModel.currency)
+      // create the accounts required for the settlementModel for existing participants
+      await ParticipantService.createAssociatedParticipantAccounts(settlementModel.currency, ledgerAccountType.ledgerAccountTypeId, trx)
```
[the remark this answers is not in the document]
vgenev: we can think about it, yeah. Although aren't we mixing some of the workflows? To create an account during transfer process is not exactly straightforward but on the other hand, if this FSP doesn't trade never in that currency, maybe is worth it.

src/models/position/facade.js
vgenev: We cannot create a participant without having default settlement model on a first place:

https://github.com/mojaloop/central-ledger/blob/e3192cd604e8f0cd2b29ee845635a4cd1e19289f/src/api/participants/handler.js#L115

I am wondering, though, isn't that a change for major version change?

mdebarros: 1. So there is no way we can ever have an error thrown here? What about upgrading from the previous version where we have participants, and currency specific settlementModels but no default exists. And a transfer is sent through that doesn't match any settlementModel?

2. Possibly. @elnyry-sam-k what are your thoughts?

vgenev: So at the moment it still throws an error, but the message is: 

`Cannot read property 'settlementAccountTypeId' of undefined` on line 62. 

Maybe we should be more specific on the error?

lewisdaly: Yes please. In my opinion, errors like the above are really nasty to see.

src/api/participants/handler.js
```diff
@@ -102,9 +102,18 @@ const create = async function (request, h) {
...
       }
     } else {
-      const participantCurrencyId1 = await ParticipantService.createParticipantCurrency(participant.participantId, request.payload.currency, ledgerAccountTypes.POSITION, false)
-      const participantCurrencyId2 = await ParticipantService.createParticipantCurrency(participant.participantId, request.payload.currency, ledgerAccountTypes.SETTLEMENT, false)
-      participant.currencyList = [await ParticipantService.getParticipantCurrencyById(participantCurrencyId1), await ParticipantService.getParticipantCurrencyById(participantCurrencyId2)]
+      const defaultSettlementModels = allSettlementModels.filter(model => model.currencyId === null)
+      if (Array.isArray(defaultSettlementModels) && defaultSettlementModels.length > 0) {
+        for (const settlementModel of defaultSettlementModels) {
+          const [participantCurrencyId1, participantCurrencyId2] = await Promise.all([
+            ParticipantService.createParticipantCurrency(participant.participantId, request.payload.currency, settlementModel.ledgerAccountTypeId, false),
```
lewisdaly, line 109: What happens if `createParticipantCurrency(...)` throws an error? I don't see a test case for that below.

src/api/participants/handler.js
```diff
@@ -102,9 +102,18 @@ const create = async function (request, h) {
         }
       }
     } else {
-      const participantCurrencyId1 = await ParticipantService.createParticipantCurrency(participant.participantId, request.payload.currency, ledgerAccountTypes.POSITION, false)
-      const participantCurrencyId2 = await ParticipantService.createParticipantCurrency(participant.participantId, request.payload.currency, ledgerAccountTypes.SETTLEMENT, false)
-      participant.currencyList = [await ParticipantService.getParticipantCurrencyById(participantCurrencyId1), await ParticipantService.getParticipantCurrencyById(participantCurrencyId2)]
+      const defaultSettlementModels = allSettlementModels.filter(model => model.currencyId === null)
```
lewisdaly, line 105: This case, where there is only 1 settlement model looks very similar to the above case, where there is more than 1 settlement model. Is there a way we can make it more generic and easier to read?

src/api/settlementModels/routes.js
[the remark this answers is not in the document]
lewisdaly: Just so I can better understand - the new behaviour is to default to the existing currency for the settlement model?

src/domain/settlement/index.js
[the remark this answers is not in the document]
lewisdaly: 2 questions:

1. This (current implementation, not Miguel's suggestion) feels like an odd side effect to me. If a user is creating a settlement model, and the ledger account type in their specified currency doesn't exist, shouldn't that be an error?

2. It also seems odd to me that by specifying the currency in this request (which users will default to because it was previously a required property) a bunch of other stuff gets created. Shouldn't the user explicitly have to create the 'associated participant accounts' themselves?

lewisdaly (changes requested):
I've mostly got some questions about how this works.... Also I don't see any integration tests.

PR #836, opened 2021-05-03
feat(db migrations): fix subid db

vijayg10:
**Issue:**
The subIdType field in quoteParty table is currently with wrong datatype and also it is referring the partyType field as foreign key which is not as per the specification.

**Fix:**
Changed the datatype of partySubIdOrTypeId in quoteParty table
Removed the foreignkey reference

**Breaking Changes:**
The datatype of the subId field in the quoteParty table is changed from integer to string.
So the upgrade should be fine, but when downgrading if a non-integer value of the subId exists in the database then it may break the downgrade process.


mdebarros (approved):
+1

PR #839, opened 2021-05-07
feat(#2123): default settlement model added

vgenev:
Added functionality to support Default Settlement Model for all participants accounts which doesn't have settlement model in the account currency


src/models/position/facade.js
```diff
@@ -52,7 +52,15 @@ const prepareChangeParticipantPositionTransaction = async (transferList) => {
     const knex = await Db.getKnex()
     const participantName = transferList[0].value.content.payload.payerFsp
     const currencyId = transferList[0].value.content.payload.amount.currency
-    const settlementModel = await SettlementModelCached.getByLedgerAccountTypeId(Enum.Accounts.LedgerAccountType.POSITION)
+    const allSettlementModels = await SettlementModelCached.getAll()
+    let settlementModels = allSettlementModels.filter(model => model.currencyId === currencyId)
+    if (!(Array.isArray(settlementModels)) || settlementModels.length <= 0) {
+      settlementModels = allSettlementModels.filter(model => model.currencyId === null) // Default settlement model
+      if (!settlementModels || (!(Array.isArray(settlementModels) || settlementModels.length <= 0))) {
+        throw ErrorHandler.Factory.createFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes.GENERIC_SETTLEMENT_ERROR, 'No Default Settlement Model has been set')
```
mdebarros, line 60: I wonder if this should be more like "Unable to find a matching or default, Settlement Model"?

src/domain/settlement/index.js
```diff
@@ -46,11 +43,14 @@ const createSettlementModel = async (settlementModel, trx = null) => {
       settlementInterchangeId, settlementDelayId, settlementModel.currency,
       settlementModel.requireLiquidityCheck,
       ledgerAccountType.ledgerAccountTypeId, settlementAccountType.ledgerAccountTypeId, settlementModel.autoPositionReset, trx)
-
-    // create the accounts required for the settlementModel for existing participants
-    await ParticipantService.createAssociatedParticipantAccounts(settlementModel.currency, ledgerAccountType.ledgerAccountTypeId, trx)
-    await ParticipantService.createAssociatedParticipantAccounts(settlementModel.currency, settlementAccountType.ledgerAccountTypeId, trx)
-
+    // Commented out because in DEFERREDNET settlement model the Limit cannot be added or adjusted after this is executed
```
mdebarros, line 46: Is there a good reason to keep this commented out code?

package.json
```diff
@@ -1,23 +1,23 @@
 {
   "name": "@mojaloop/central-ledger",
-  "version": "13.4.0",
+  "version": "13.5.0",
```
mdebarros, line 3: How certain are we that there no breaking change here when upgrading from 13.4.0 to 13.5.0?

I can't think of anything here that is breaking. If we have any doubts we should try upgrading from 13.4.0 to 13.5.0 to see if there are any issues.

mdebarros (commented):
See my comments.

src/domain/settlement/index.js
```diff
@@ -28,15 +28,12 @@
 const SettlementModelModel = require('../../models/settlement/settlementModel')
 const LedgerAccountTypeModel = require('../../models/ledgerAccountType/ledgerAccountType')
 const Enum = require('@mojaloop/central-services-shared').Enum.Settlements
-const ParticipantService = require('../participant')
+// const ParticipantService = require('../participant')
```
lewisdaly, line 31: let's remove commented code

src/models/participant/facade.js
```diff
@@ -351,7 +351,7 @@ const getParticipantLimitByParticipantCurrencyLimit = async (participantId, curr
           .innerJoin('participantCurrency AS pc', 'pc.participantId', 'participant.participantId')
           .innerJoin('participantLimit AS pl', 'pl.participantCurrencyId', 'pc.participantCurrencyId')
           .select(
-            'participant.participantID AS participantId',
+            'participant.participantId AS participantId',
```
lewisdaly, line 354: how does this work? I thought mysql was case sensitive.

src/api/participants/handler.js
```diff
@@ -93,18 +93,30 @@ const create = async function (request, h) {
     const settlementModels = allSettlementModels.filter(model => model.currencyId === request.payload.currency)
     if (Array.isArray(settlementModels) && settlementModels.length > 0) {
       for (const settlementModel of settlementModels) {
-        const participantCurrencyId1 = await ParticipantService.createParticipantCurrency(participant.participantId, request.payload.currency, settlementModel.ledgerAccountTypeId, false)
-        const participantCurrencyId2 = await ParticipantService.createParticipantCurrency(participant.participantId, request.payload.currency, settlementModel.settlementAccountTypeId, false)
+        const [participantCurrencyId1, participantCurrencyId2] = await Promise.all([
```
lewisdaly, line 96: I find this whole section from 96 -> 119 impossible to read. 

1. We are nesting 3 else blocks deep - why not `return` early on line `104` before the else block?
2. There is a lot of copying and pasting going on - line 96 -> 103 is almost the same as 108 -> 116. Let's make some generic functions to do this work and make the code more manageable
3. The error case where no default settlement model is set on line `118` is hidden at the bottom with a bunch of else statements. As a reader, there is no easy way for me to understand the conditions that would lead to triggering this error.

src/models/position/facade.js
```diff
@@ -52,7 +52,15 @@ const prepareChangeParticipantPositionTransaction = async (transferList) => {
     const knex = await Db.getKnex()
     const participantName = transferList[0].value.content.payload.payerFsp
     const currencyId = transferList[0].value.content.payload.amount.currency
-    const settlementModel = await SettlementModelCached.getByLedgerAccountTypeId(Enum.Accounts.LedgerAccountType.POSITION)
+    const allSettlementModels = await SettlementModelCached.getAll()
+    let settlementModels = allSettlementModels.filter(model => model.currencyId === currencyId)
+    if (!(Array.isArray(settlementModels)) || settlementModels.length <= 0) {
+      settlementModels = allSettlementModels.filter(model => model.currencyId === null) // Default settlement model
+      if (!settlementModels || (!(Array.isArray(settlementModels) || settlementModels.length <= 0))) {
```
lewisdaly, line 59: is `Array.isArray` really required? If we're calling `.filter` in `58`, surely this is an array, right?

lewisdaly (changes requested):
Please see comments - mostly my concern is readability.

src/models/participant/facade.js
vgenev: I wonder how the previous was working actually because of the same reason...

src/models/position/facade.js
vgenev: I agree

src/models/position/facade.js
vgenev: I agree

mdebarros (changes requested):

src/models/position/facade.js
```diff
@@ -52,7 +52,15 @@ const prepareChangeParticipantPositionTransaction = async (transferList) => {
...
+        throw ErrorHandler.Factory.createFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes.GENERIC_SETTLEMENT_ERROR, 'No Default Settlement Model has been set')
+      }
```
mdebarros, line 61: What happens if there is "no default" or matching settlement model here? :D

mdebarros (approved):
+1

src/api/participants/handler.js
```diff
@@ -90,21 +90,22 @@ const create = async function (request, h) {
     }
     const ledgerAccountIds = Util.transpose(ledgerAccountTypes)
     const allSettlementModels = await SettlementService.getAll()
-    const settlementModels = allSettlementModels.filter(model => model.currencyId === request.payload.currency)
-    if (Array.isArray(settlementModels) && settlementModels.length > 0) {
+    let settlementModels = allSettlementModels.filter(model => model.currencyId === request.payload.currency)
```
lewisdaly, line 93: This is much more readable! Thanks.

Maybe I misunderstood the original purpose, but it looks like if there are settlement models found on line `93` we never do anything with them (e.g. line 99). Is there meant to be another  closed bracket on line `99`?

src/models/participant/facade.js
lewisdaly: Hmm. If we are wrong here, where will this error be picked up? In end to end tests?

lewisdaly (dismissed):
One more question about `participants/handler.js`, but looks good to me

src/api/participants/handler.js
vgenev: ah you are right!

mdebarros (approved):
+1

PR #852, opened 2021-06-18
feat(#2300): add logging statements for all errors

kleyow:
Addresses https://github.com/mojaloop/project/issues/2300 ? 

With not much to go on from the story, this is all that I could conceivably imagine as a task for the story. 
Any direction is appreciated.

### DIFF
--- a/src/api/ledgerAccountTypes/handler.js
+++ b/src/api/ledgerAccountTypes/handler.js
@@ -26,6 +26,7 @@
 
 const LedgerAccountTypesService = require('../../domain/ledgerAccountTypes')
 const ErrorHandler = require('@mojaloop/central-services-error-handling')
+const Logger = require('@mojaloop/central-services-logger')
 
 const getAll = async function () {
   return LedgerAccountTypesService.getAll()
@@ -41,6 +42,7 @@ async function create (request, h) {
       return h.response().code(201)
     }
   } catch (err) {
+    Logger.isErrorEnabled && Logger.error(err)
     throw ErrorHandler.Factory.reformatFSPIOPError(err)
   }
 }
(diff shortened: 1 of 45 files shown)


mdebarros (commented):
I feel like this will print unnecessary errors to the log file.

Should we perhaps only print errors at the "API" (Rest API calls) or "Handler" (Kafka Consumers) application levels?

i.e.
- src/api
- src/handlers

kleyow:
@mdebarros Sounds good to me, reverted some changes. We can always take an incremental approach. Commit push

mdebarros:
> @mdebarros Sounds good to me, reverted some changes. We can always take an incremental approach. Commit pushed.

Thanks! It looks good to me.

There is something else we could change with regard to this PR --> https://github.com/mojaloop/central-ledger/blob/master/src/lib/requestLogger.js

This currently logs all incoming HapiJS requests/responses as debug log entries. I personally feel that it's too verbose to switch these to info since we already have the "audits" for ingress and egress using the Event-SDK.

Let me know your thoughts here.

kleyow:
No strong thoughts. Sounds logical enough for me. I don't have much insight of the pov of a system admin overlooking the central ledger. made the switch to info.

lewisdaly:
> Let me know your thoughts here.

I agree @mdebarros  - yeah having lots of noise on the `info` log entry makes debugging from the command line pretty tricky

kleyow:
I misread your comment. I'm a bit confused. I haven't changed `requestLogger` away from debug in my earlier commits so I feel like I'm missing some context...

mdebarros (approved):
+1

mdebarros:
> I misread your comment. I'm a bit confused. I haven't changed `requestLogger` away from debug in my earlier commits so I feel like I'm missing some context...

Exactly, you have not changed it. I was raising the question if we should, but after more though I really think we shouldn't. ;)

PR #853, opened 2021-06-22
fix: helm release 12.1 for default settlement functionality 

vgenev:
* Added new column settlementModelId to settlementWindowContent as foreign key to settlementModel


migrations/950107_settlementWindowContent_foreign_settlementModel.js
```diff
@@ -0,0 +1,46 @@
...
+              t.foreign('SettlementModelId').references('SettlementModelId').inTable('settlementModel')
+            })
+          }
+        })
+    }
+  })
+}
+
+exports.down = function (knex) {
+  return knex.schema.dropTableIfExists('settlementWindowContent')
```
mdebarros, line 45: This should remove the key constraint and the column, not drop the table.

mdebarros (changes requested):
See my comment.

migrations/950107_settlementWindowContent_foreign_settlementModel.js
vgenev: oh.. ok..

mdebarros (approved):
+1

PR #869, opened 2021-11-08
fix(#2557): error notification to payer fsp, header for source having wrong value

mdebarros:
- fix for https://github.com/mojaloop/project/issues/2557
- Updated fulfil-handler logic to produce an action type "ABORT_VALIDATION" when a validation error occurs (i.e. invalid fulfilment)
- updated util/scripts/populateTestData.sh script to create a default settlement model for local dev testing
- chore: updated dependencies
    - main update to central-services-shared
    - added get-port to ncurc ignore list due to the breaking changes it introduces
    - updated all other dependencies
    - fixed audit-resolve issues


lewisdaly:
the mpath vulnerability is marked as critical, so we probably shouldn't ignore it.

This PR: https://github.com/mojaloop/object-store-lib/pull/33 will fix it

mdebarros:
> the mpath vulnerability is marked as critical, so we probably shouldn't ignore it.
> 
> This PR: [mojaloop/object-store-lib#33](https://github.com/mojaloop/object-store-lib/pull/33) will fix it

I have a bigger concern, I am not sure anybody is even using bulk functionality, and thus its not being maintained.

lewisdaly:
> the mpath vulnerability is marked as critical, so we probably shouldn't ignore it

Yeah I can cover this in an upcoming PR

> I have a bigger concern, I am not sure anybody is even using bulk functionality, and thus it's not being maintained.

Yeah that's true as well... but don't we have tests for it?

lewisdaly:
Also, how did this not break any tests?

lewisdaly (approved):

mdebarros:
> > the mpath vulnerability is marked as critical, so we probably shouldn't ignore it
> 
> Yeah I can cover this in an upcoming PR
> 
> > I have a bigger concern, I am not sure anybody is even using bulk functionality, and thus it's not being maintained.
> 
> Yeah that's true as well... but don't we have tests for it?

Because bulk is in "incubation"

PR #901, opened 2022-05-20
fix: error codes for liquidity and ndc limit check

vijayg10:


src/models/position/facade.js
```diff
@@ -183,11 +183,11 @@ const prepareChangeParticipantPositionTransaction = async (transferList) => {
           const { transfer, transferState, rawMessage, transferAmount } = reservedTransfers[transferId]
           if (new MLNumber(availablePositionBasedOnLiquidityCover).toNumber() < transferAmount.toNumber()) {
             transferState.transferStateId = Enum.Transfers.TransferInternalState.ABORTED_REJECTED
-            transferState.reason = ErrorHandler.Enums.FSPIOPErrorCodes.PAYER_FSP_INSUFFICIENT_LIQUIDITY.message
+            transferState.reason = 'PAYER_FSP_INSUFFICIENT_LIQUIDITY'
```
mdebarros, line 186: Please add missing FSPIOPErrorCodes enums to this library --> https://github.com/mojaloop/central-services-error-handling/blob/master/src/enums.js#L78.

Same comment applies to the reason below.

src/handlers/positions/handler.js
```diff
@@ -163,7 +163,7 @@ const positions = async (error, messages) => {
           return true
         } else {
           Logger.isInfoEnabled && Logger.info(Utility.breadcrumb(location, `payerNotifyInsufficientLiquidity--${actionLetter}2`))
-          const fspiopError = ErrorHandler.Factory.createFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes.PAYER_FSP_INSUFFICIENT_LIQUIDITY)
+          const fspiopError = ErrorHandler.Factory.createFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes[transferState.reason] || ErrorHandler.Enums.FSPIOPErrorCodes.SERVER_ERROR)
```
mdebarros, line 166: Why are we doing the `||` part?

mdebarros (changes requested):

src/models/position/facade.js
vijayg10: I refactored the fix with a different approach

mdebarros (approved):

PR #911, opened 2022-07-22
feat: add bulk error handling notification callbacks

kleyow:
image


src/handlers/bulk/prepare/handler.js
```diff
@@ -104,16 +105,23 @@ const bulkPrepare = async (error, messages) => {
...
 
     const { hasDuplicateId, hasDuplicateHash } = await Comparators.duplicateCheckComparator(bulkTransferId, payload, BulkTransferService.getBulkTransferDuplicateCheck, BulkTransferService.saveBulkTransferDuplicateCheck)
-    if (hasDuplicateId && hasDuplicateHash) { // TODO: handle resend :: GET /bulkTransfer
-      Logger.isInfoEnabled && Logger.info(Util.breadcrumb(location, `resend--${actionLetter}1`))
-      Logger.isErrorEnabled && Logger.error(Util.breadcrumb(location, 'notImplemented'))
-      return true
+    if (hasDuplicateId && hasDuplicateHash) {
+      Logger.isErrorEnabled && Logger.error(Util.breadcrumb(location, `callbackErrorModified--${actionLetter}1`))
+
+      const fspiopError = ErrorHandler.Factory.createFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes.MODIFIED_REQUEST, 'Bulk transfer prepare duplicate')
```
elnyry-sam-k, line 111: The behavior on duplicate requests is that:
- If the ID (here bulkTransferID) and the message hash matches, it is essentially a "resend" so in this case - "IF the bulkTransfer is in a final state, we treat it as a GET /bulkTransfers/{ID} call and send out a PUT /bulkTransfers/{ID} with the same payload as that of a final notification. If however, the bulkTransfer is in a non-final state, then we do not send anything back, allowing the final notification to be sent once processing is done.
- On a bulkTransfer if a previously used ID is used but the message is different or modified (hash is different), then we send out the error callback saying its a "modified request"

Here's the section in the FSPIOP API spec discussing this behavior: https://docs.mojaloop.io/api/fspiop/v1.1/api-definition.html#duplicate-analysis-in-server-on-receiving-a-http-post-request

src/handlers/bulk/prepare/handler.js
```diff
@@ -106,17 +107,70 @@ const bulkPrepare = async (error, messages) => {
 
     const { hasDuplicateId, hasDuplicateHash } = await Comparators.duplicateCheckComparator(bulkTransferId, payload, BulkTransferService.getBulkTransferDuplicateCheck, BulkTransferService.saveBulkTransferDuplicateCheck)
     if (hasDuplicateId && hasDuplicateHash) {
-      Logger.isErrorEnabled && Logger.error(Util.breadcrumb(location, `callbackErrorModified--${actionLetter}1`))
+      const eventDetail = { functionality: Enum.Events.Event.Type.NOTIFICATION, action: TransferEventAction.PREPARE_DUPLICATE }
+
+      // Check to see if the resend is coming from a participant of involved
+      // with the bulk transfer to avoid potential data leakage.
+      const participants = await BulkTransferService.getParticipantsById(bulkTransferId)
+      if (![participants.payeeFsp, participants.payerFsp].includes(message.value.from)) {
```
elnyry-sam-k, line 115: This one and the next check should be unreachable code, isn't it? "hasDuplicateHash" should indicate that the headers and payload are exactly same as before.. So the check that this is coming from the same Payer FSP as before should already be covered there..

src/handlers/bulk/prepare/handler.js
```diff
@@ -106,17 +107,70 @@ const bulkPrepare = async (error, messages) => {
...
+        await Kafka.proceed(Config.KAFKA_CONFIG, params, { consumerCommit, fspiopError: fspiopError.toApiErrorObject(Config.ERROR_HANDLING), eventDetail, fromSwitch })
+        throw fspiopError
+      }
 
-      const fspiopError = ErrorHandler.Factory.createFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes.MODIFIED_REQUEST, 'Bulk transfer prepare duplicate')
-      const eventDetail = { functionality: Enum.Events.Event.Type.NOTIFICATION, action }
-      params.message.value.content.uriParams = { id: bulkTransferId }
+      // Resend should typically only occur by the Payer FSP, but lets
+      // handle if the same request is somehow sent by the Payer FSP.
+      const isPayeeRequest = participants.payeeFsp === message.value.from
```
elnyry-sam-k, line 125: shouldn't be needed, isn't it? (handled by "hasDuplicateHash")

elnyry-sam-k (approved):

kleyow:
https://github.com/mojaloop/central-ledger/pull/913

kleyow:
https://github.com/mojaloop/project/issues/2798

PR #916, opened 2022-08-11
feat(mojaloop/#2801): add fulfil timestamp validation and more error handling

kleyow:
feat(mojaloop/#2801): add fulfil timestamp validation and more error handling - [https://github.com/mojaloop/project/issues/2801](https://github.com/mojaloop/project/issues/2801)

- Port timestamp validation logic that was previously in the `bulk-api-adapter`
- Add fulfil handler logic to invalidate individualTransfers in a bulkTransfer
- Update bulk-processsing handler to handle new validation failure state


config/default.json
```diff
@@ -18,6 +18,7 @@
     "CREATE_RETRY_INTERVAL_MILLIS": 200,
     "DEBUG": false
   },
+  "MAX_FULFIL_TIMEOUT_DURATION_SECONDS": 300,
```
mdebarros, line 21: I believe this is a breaking change right?

I.e. What is the impact if we do not provide the `MAX_FULFIL_TIMEOUT_DURATION_SECONDS` config either via default.json or env file?

mdebarros (changes requested):

config/default.json
kleyow: There is a default timeout that it will default to https://github.com/mojaloop/central-ledger/pull/916/files#diff-b69d6bde1ed95b38264df8b79c66099602625f0eb541d4e98001baed223096d7R213

src/handlers/bulk/shared/validator.js
```diff
@@ -203,9 +205,34 @@ const validateBulkTransferFulfilment = async (payload, headers) => {
     return { isValid, reasons }
   }
   isValid = isValid && await validateFspiopSourceAndDestination(payload, headers)
+  isValid = isValid && validateCompletedTimestamp(payload)
   return { isValid, reasons }
 }
 
+const validateCompletedTimestamp = (payload) => {
+  const maxLag = (Config.MAX_FULFIL_TIMEOUT_DURATION_SECONDS || defaultLagSeconds) * 1000
```
mdebarros, line 213: Would it not be better to have the default values set in the config.js instead of here?

src/lib/config.js
```diff
@@ -3,6 +3,7 @@ const RC = require('rc')('CLEDG', require('../../config/default.json'))
 module.exports = {
   HOSTNAME: RC.HOSTNAME.replace(/\/$/, ''),
   PORT: RC.PORT,
+  MAX_FULFIL_TIMEOUT_DURATION_SECONDS: RC.MAX_FULFIL_TIMEOUT_DURATION_SECONDS,
```
mdebarros, line 6: I.e. I feel that it would be better to have the default value set here.

kleyow: Yes. It's better. done.

src/models/bulkTransfer/facade.js
```diff
@@ -172,10 +172,54 @@ const saveBulkTransferErrorProcessing = async (payload, stateReason = null, isVa
...
 }
 
+const saveBulkTransferAborting = async (payload, stateReason = null) => {
+  try {
+    const bulkTransferFulfilmentRecord = {
+      bulkTransferId: payload.bulkTransferId,
+      completedDate: Time.getUTCString(new Date(payload.completedTimestamp))
+    }
+
+    const state = 'ABORTING'
```
mdebarros, line 182: Please use enum.

mdebarros (changes requested):
LGTM, but minor comment for you to address.

mdebarros (approved):
+1

PR #934, opened 2023-01-12
feat(mojaloop/3050): implemented settlement model rules engine

vijayg10:
feat([mojaloop/3050](https://github.com/mojaloop/project/issues/3050)): implemented settlement model rules engine


kleyow (commented):
LGTM so far.

sri-miriyala:
@vijayg10  I see one unit test for default settlement model. Do we also need to add a test for remittance model? Also the config/rules-settlement-model.json only has an entry for default settlement model. Don't we also need an entry for remittance model?

PR #935, opened 2023-01-13
feat(mojaloop/3051): integrate rule engine

vijayg10:
feat([mojaloop/3051](https://github.com/mojaloop/project/issues/3051)): integrate rule engine


src/handlers/transfers/handler.js
```diff
@@ -212,7 +228,7 @@ const prepare = async (error, messages) => {
         Logger.isErrorEnabled && Logger.error(Util.breadcrumb(location, { path: 'validationFailed' }))
         try {
           Logger.isInfoEnabled && Logger.info(Util.breadcrumb(location, 'saveInvalidRequest'))
-          await TransferService.prepare(payload, reasons.toString(), false)
+          await TransferService.prepare(message, reasons.toString(), false)
```
kleyow, line 231: I think I missed this in my commit Vijay. 
Hmm...I guess we might have to fetch the settlement model regardless of validation since we save the transfer even if it failed validation.

kleyow (changes requested):
Looking good so far aside from test work. 
1 important comment. @vijayg10

kleyow:
I fixed some of the coverage, linting and deps on Sri's work and pulled them in here. 
I briefly looked into why integration tests are failing.

It seems theres something going on with the Kafka connect/disconnect between the transfers/handlers.test.js and rules/settlement-models-rules.engine.js. 

They seem to work fine when they run independently but not after each other.

src/handlers/transfers/handler.js
```diff
@@ -177,21 +187,44 @@ const prepare = async (error, messages) => {
...
+      // Select settlement model here
+      let settlementModel
+      try {
+        const allSettlementModels = await SettlementModelCached.getAll()
+        const settlementModels = allSettlementModels.filter(model => !model.currencyId || model.currencyId === payload.amount.currency)
+        if (settlementModels.length === 0) {
+          throw ErrorHandler.Factory.createFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes.GENERIC_SETTLEMENT_ERROR, 'Unable to find a matching or default, Settlement Model')
         }
+        settlementModel = settlementModels.find(sm => sm.ledgerAccountTypeId === Enum.Accounts.LedgerAccountType.POSITION)
+        if (Config.INCLUDE_DECODED_TRANSACTION_OBJECT) {
```
kleyow, line 199: I don't necessarily think we need this config option anymore.

vijayg10: Done

.circleci/config.yml
```diff
@@ -815,11 +815,12 @@ workflows:
           context: org-global
           requires:
             - setup
-            - test-dependencies
+            ## TODO: Commenting the test steps for snapshot release temporarily
```
kleyow, line 818: You can remove these now.

kleyow (approved):
One comment, but lgtm.

.circleci/config.yml
vijayg10: Hi @kleyow , lets keep this for feature branch so that we can make snapshot releases. We can remove this when we merge this to master

PR #936, opened 2023-01-16
feat(mojaloop/#3054): add option to include decoded transaction in messages

kleyow:
feat(mojaloop/#3054): add option to include decoded transaction in messages- https://github.com/mojaloop/project/issues/3054

- Add configurable functionality to include decoded transaction object in kafka messages in the prepare handler
- Update tests (ilpPacket in previous tests was invalid, test updated with a valid transfer from test enviornment)
- Update dependencies
- Peg knex version @ 2.3.0 due to issue with it seeding/migrating the database differently and leading to integration test failure. Unsure of source of failure, only preliminary investigation


mdebarros:
Improvements for tests:
1. Happy path add check of the actual transaction object and compare the contents to the test-data's decoded ILP-Packet
2. Add unhappy Path for when the ILP-Packet is invalid (i.e. add validation logic by using the ILP packets validation function)

mdebarros (approved):
+1

PR #965, opened 2023-08-30
feat(mojaloop/#3498): alter message keys for prepare, fulfil and timeout

kleyow:
feat(mojaloop/#3498): alter message keys for prepare, fulfil and timeout - https://github.com/mojaloop/project/issues/3498

- Refator `central-services-shared` proceed to lib with simplified keying logic
- Remove keying from some `NOTIFICATION` messages in transfers handler
- Keyed successful position prepares by payer account id
- Keyed successful position fulfils by payee account id
- Keyed all position aborts by payer account id


src/handlers/transfers/handler.js
```diff
@@ -198,7 +198,9 @@ const prepare = async (error, messages) => {
         Logger.isInfoEnabled && Logger.info(Util.breadcrumb(location, `positionTopic1--${actionLetter}7`))
         functionality = TransferEventType.POSITION
         const eventDetail = { functionality, action }
-        await Kafka.proceed(Config.KAFKA_CONFIG, params, { consumerCommit, eventDetail, toDestination })
+        // Key position prepare message with payer account id
+        const payerAccount = await Participant.getAccountByNameAndCurrency(payload.payerFsp, payload.amount.currency, Enum.Accounts.LedgerAccountType.POSITION)
```
vijayg10, line 202: For this particular case here(happy path), if we want to save a DB query, the participant account id is already fetched in [saveTransferPrepared](https://github.com/mojaloop/central-ledger/blob/c8e0d57776bfcd055df7fe981d88f8e2ccce258c/src/models/transfer/facade.js#L414) of facade. You can pass it down to access here. But as long as the query is cached, I don't think its a must. @mdebarros What do you suggest?

src/handlers/timeouts/handler.js
```diff
@@ -96,7 +96,8 @@ const timeout = async () => {
           } else if (result[i].transferStateId === Enum.Transfers.TransferInternalState.RESERVED_TIMEOUT) {
             message.metadata.event.type = Enum.Events.Event.Type.POSITION
             message.metadata.event.action = Enum.Events.Event.Action.TIMEOUT_RESERVED
-            await Kafka.produceGeneralMessage(Config.KAFKA_CONFIG, Producer, Enum.Kafka.Topics.POSITION, Enum.Events.Event.Action.TIMEOUT_RESERVED, message, state, result[i].payerFsp, span)
+            // Key position timeouts with payer account id
+            await Kafka.produceGeneralMessage(Config.KAFKA_CONFIG, Producer, Enum.Kafka.Topics.POSITION, Enum.Events.Event.Action.TIMEOUT_RESERVED, message, state, result[i].payerParticipantId?.toString(), span)
```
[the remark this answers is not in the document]
kleyow: To my knowledge, yes. https://github.com/mojaloop/central-ledger/blob/main/src/models/transfer/facade.js#L718
There is an integration test for timeout as well that asserts the key https://github.com/mojaloop/central-ledger/pull/965/files#diff-de162bcc4f2b73f73b420c6022a5a81a75bf8a1554dd3cbf4998b7d2e394f768R1305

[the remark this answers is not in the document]
vijayg10: Ok

src/handlers/transfers/handler.js
mdebarros: Lets create a story for this enhancement.

This in practice shouldn't be an issue as long as caching is enabled, but thats not too say we shouldn't make such improvements going forward. I also think it would be good to address it as a dedicated story as there may be other such opportunities to reduce "potential" duplicate queries.

kleyow: I've created a story to optimize duplicated queries for participant currency id with a clause to investigate other instances of duplicated queries (still updating the story, just a placeholder) https://github.com/mojaloop/project/issues/3515

mdebarros (approved):
+1

PR #967, opened 2023-09-06
chore(mojaloop/#3518): update database lib

oderayi:
chore(mojaloop/#3518):update database lib - https://github.com/mojaloop/project/issues/3518
- replaced `central-services-database` with `database-lib`


mdebarros (approved):
+1

PR #1006, opened 2024-04-18
feat(mojaloop/#3844): added corner cases impl. for FX; added unit-tests

geka-evk:
feat(mojaloop/#3844): added corner cases impl. for FX; added unit-tests


vijayg10 (commented):
Looks good to me except logger refactoring. It causes unnecessary template rendering as we previously discussed once. I am not sure the performance impact of it, we can discuss with the team and decide on that. Found some TODO comments, I assume you are planning to discuss those and resolve.

vijayg10 (approved):

PR #1008, opened 2024-04-22
chore(mojaloop/#3819): add error for payee initiated fx conversions

kleyow:
chore(mojaloop/#3819): add error for payee initiated fx conversions - https://github.com/mojaloop/project/issues/3819

### DIFF
--- a/src/domain/fx/cyril.js
+++ b/src/domain/fx/cyril.js
@@ -25,6 +25,7 @@
 
 const Metrics = require('@mojaloop/central-services-metrics')
 const { Enum } = require('@mojaloop/central-services-shared')
+const ErrorHandler = require('@mojaloop/central-services-error-handling')
 const TransferModel = require('../../models/transfer/transfer')
 const ParticipantFacade = require('../../models/participant/facade')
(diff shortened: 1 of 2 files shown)


src/domain/fx/cyril.js
```diff
@@ -82,20 +83,23 @@ const getParticipantAndCurrencyForFxTransferMessage = async (payload) => {
...
+      'Payee FX conversion not implemented'
+    )
+    // participantName = payload.counterPartyFsp
+    // currencyId = payload.targetAmount.currency
+    // amount = payload.targetAmount.amount
+    // await watchList.addToWatchList({
+    //   commitRequestId: payload.commitRequestId,
+    //   determiningTransferId: payload.determiningTransferId,
+    //   fxTransferTypeId: Enum.Fx.FxTransferType.PAYEE_CONVERSION
+    // })
```
vijayg10, line 102: I think payee side currency conversion is already implemented by seeing this code. I don't think we need to disable that.

kleyow: But then we would need to give it the whole suite of testing, no? I thought testing payee conversion was out of scope? @vijayg10

PR #1035, opened 2024-05-17
feat(mojaloop/#3904): add position event fx timeout reserved batch handling

kleyow:
feat(mojaloop/#3904): add position event fx timeout reserved batch handling


src/domain/position/binProcessor.js
```diff
@@ -88,6 +89,13 @@ const processBins = async (bins, trx) => {
     Enum.Accounts.LedgerEntryType.PRINCIPLE_VALUE
   )
 
+  const latestInitiatingFxTransferInfoByFxCommitRequestId = await BatchPositionModel.getFxTransferInfoList(
+    trx,
+    commitRequestIdList,
+    Enum.Accounts.TransferParticipantRoleType.INITIATING_FSP,
+    Enum.Accounts.LedgerEntryType.PRINCIPLE_VALUE
+  )
+
```
vijayg10, line 98: I think, the fxTransferParticipant table logic is messed up here https://github.com/mojaloop/central-ledger/blob/0eea335ebe6a04fd0f02ec7d2b8f83ba61b530e8/src/models/fxTransfer/fxTransfer.js#L146
We can't say initiatingFsp is always related to sourceAmount and currency. It depends on the use case.

So, I think its better to get the amount values from participantPositionChange. i.e, findout the fxTransferStateChanges for a commitRequestI, then find out participantPositionChanges for those fxTransferStateChanges filtered by the participant accountID (kafka message key).

kleyow: @vijayg10 done. take another look when you can.

src/models/position/batch.js
```diff
@@ -211,6 +211,63 @@ const getTransferByIdsForReserve = async (trx, transferIds) => {
...
+        'fxTransferStateChange.commitRequestId AS commitRequestId',
+        'pp.participantCurrencyId AS participantCurrencyId'
+      )
+    const info = {}
+    for (const participantPositionChange of participantPositionChanges) {
+      if (!(participantPositionChange.commitRequestId in info)) {
+        info[participantPositionChange.commitRequestId] = {}
+      }
+      if (participantPositionChange.participantCurrencyId) {
+        info[participantPositionChange.commitRequestId][participantPositionChange.participantCurrencyId] = participantPositionChange
```
vijayg10, line 261: This logic seems to work if we assume there will be only one participantPositionChange per participantCurrencyId for a given commitRequestId. Although we don't have any use case to have multiple, but is it safe to assume so in the model level?

src/domain/position/binProcessor.js
```diff
@@ -88,6 +89,15 @@ const processBins = async (bins, trx) => {
...
   )
 
+  // Fetch all RESERVED participantPositionChanges associated with a commitRequestId
+  // These will contain the value that was reserved for the fxTransfer
+  // We will use these values to revert the position on timeouts
+  const fetchedReservedPositionChangesByCommitRequestIds =
+    await BatchPositionModel.getReservedPositionChangesByCommitRequestIds(
+      trx,
+      commitRequestIdList
+    )
```
vijayg10, line 99: Do we need to get position changes for all the accountIds? Or do you want to pass the accountId and query only for that accountId?

src/domain/position/fx-timeout-reserved.js
```diff
@@ -0,0 +1,153 @@
...
+  // Position action FX_RESERVED_TIMEOUT event messages are keyed with payer account id.
+  // We need to revert the payer's position for the source currency amount of the fxTransfer.
+  // We need to notify the payee of the timeout.
+  if (fxTimeoutReservedBins && fxTimeoutReservedBins.length > 0) {
+    for (const binItem of fxTimeoutReservedBins) {
+      Logger.isDebugEnabled && Logger.debug(`processPositionFxTimeoutReservedBin::binItem: ${JSON.stringify(binItem.message.value)}`)
+      const participantAccountId = binItem.message.key.toString()
+      const commitRequestId = binItem.message.value.content.uriParams.id
+      const fxp = binItem.message.value.to
+      const payerFsp = binItem.message.value.from
```
vijayg10, line 42: Can we rename fxp as counterPartyFsp and payerFsp as initiatingFsp? Because 'from' is not always payer, it can be payee in payee side currency conversion.

src/domain/position/fx-timeout-reserved.js
```diff
@@ -0,0 +1,153 @@
...
+
+      // If the transfer is not in `RESERVED_TIMEOUT`, a position fx-timeout-reserved message was incorrectly published.
+      // i.e Something has gone extremely wrong.
+      if (accumulatedFxTransferStates[commitRequestId] !== Enum.Transfers.TransferInternalState.RESERVED_TIMEOUT) {
+        throw ErrorHandler.Factory.createInternalServerFSPIOPError(ErrorHandler.Enums.FSPIOPErrorCodes.INTERNAL_SERVER_ERROR.message)
+      } else {
+        Logger.isDebugEnabled && Logger.debug(`accumulatedFxTransferStates: ${JSON.stringify(accumulatedFxTransferStates)}`)
+
+        const transferAmount = fetchedReservedPositionChangesByCommitRequestIds[commitRequestId][participantAccountId].value
```
vijayg10, line 51: As per my earlier comment, we need to make sure if we get single participantPositionChange item always. Otherwise we need to get the list of position changes and aggregate the resultant position value to be reverted.

kleyow: Makes sense, but do we ever forsee a operation doing a RESERVED transfer state change doing having multiple participant position changes?

src/domain/position/fx-timeout-reserved.js
kleyow: Yep! Mb. done.

src/domain/position/binProcessor.js
kleyow: We model a lot of our reads to just do one select so we have all the data at hand in the bin processor, no?
Querying specific accountId's one at a time would defeat the purpose.

src/models/position/batch.js
kleyow: Yes it is assuming that there is 1 RESERVED type participant position change per commitRequestId per account.
Yea...it doesn't feel right. Where do you propose we validate such assumptions?

vijayg10: I think, lets leave that assumption as an elaborated comment for now.

src/domain/position/fx-timeout-reserved.js
vijayg10: Yeah I can't see any use case atm. Lets add a comment illustrating that issue as per my other comment.

src/models/position/batch.js
kleyow: done. added a comment on the model function.

src/domain/position/binProcessor.js
vijayg10: Oh you are right. I thought we only work on single accountId in the bin processor.

vijayg10 (approved):

PR #1054, opened 2024-06-18
fix: avoid 2 extra db requests

kalinkrustev:


src/handlers/transfers/FxFulfilService.js
```diff
@@ -307,17 +307,19 @@ class FxFulfilService {
 
   async processFxFulfil({ transfer, payload, action }) {
     await this.FxTransferModel.fxTransfer.saveFxFulfilResponse(transfer.commitRequestId, payload, action)
-    const cyrilOutput = await this.cyril.processFxFulfilMessage(transfer.commitRequestId, payload)
+    if (!transfer.fxWatchListId) {
+      throw new Error(`Commit request ID ${transfer.commitRequestId} not found in watch list`)
+    }
```
vijayg10, line 312: Can we still have a function in cyril 'processFxFulfilMessage' for this check? I mean. you can remove the call 'getAllDetailsByCommitRequestId' in that function, but I would recommend the queries (getItemInWatchListByCommitRequestId) on watchlist in the cyril functions instead of having a join in other place. Because the watchlist logic may change and I wouldn't suggest merging watchList in the places other than cyril. I know it's additional query but I hope it will not impact performance that much compared to left join.

src/models/fxTransfer/fxTransfer.js
```diff
@@ -90,9 +90,11 @@ const getAllDetailsByCommitRequestId = async (commitRequestId) => {
         .leftJoin('fxTransferStateChange AS tsc', 'tsc.commitRequestId', 'fxTransfer.commitRequestId')
         .leftJoin('transferState AS ts', 'ts.transferStateId', 'tsc.transferStateId')
         .leftJoin('fxTransferFulfilment AS tf', 'tf.commitRequestId', 'fxTransfer.commitRequestId')
+        .leftJoin('fxWatchList AS wl', 'wl.commitRequestId', 'fxTransfer.commitRequestId')
```
vijayg10, line 93: Please see my above comment.

vijayg10 (changes requested):

PR #1059, opened 2024-07-04
feat(mojaloop/csi-190): add new state and functionality to handle proxied transfers

kleyow:


src/models/transfer/facade.js
```diff
@@ -1088,6 +1090,21 @@ const transferStateAndPositionUpdate = async function (param1, enums, trx = null
...
+        transferId,
+        transferStateId: TransferInternalState.RESERVED_FORWARDED,
+        reason: null,
+        createdDate: Time.getUTCString(new Date())
+      })
+  } catch (err) {
+    throw ErrorHandler.Factory.reformatFSPIOPError(err)
+  }
+}
+
```
kalinkrustev, line 1107: I do not know very well the state update engine, but this looks a bit unexpected approach to me. My comment is not directly related to the feature addressed by this PR, but what gives us a guarantee that another insert for a different state is not happening at the same time resulting in both inserts being successful?

PR #1060, opened 2024-07-16
feat(csi-22): add proxy lib to handlers

kleyow:


src/lib/proxyCache.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+const { createProxyCache } = require('@mojaloop/inter-scheme-proxy-cache-lib')
+const Config = require('./config.js')
+const ParticipantService = require('../../src/domain/participant')
+
+const proxyCache = createProxyCache(
+  Config.PROXY_CACHE_CONFIG.type,
+  Config.PROXY_CACHE_CONFIG.proxyConfig
+)
+const getCache = () => {
```
geka-evk, line 10: What is the point of having this function?

src/lib/proxyCache.js
```diff
@@ -0,0 +1,33 @@
...
+  const participant = await ParticipantService.getByName(dfspId)
+  return {
+    inScheme: !!participant,
+    proxyId: !participant ? await getCache().lookupProxyByDfspId(dfspId) : null
+  }
+}
+
+const checkSameCreditorDebtorProxy = async (debtorDfspId, creditorDfspId) => {
+  const debtorProxyId = await getCache().lookupProxyByDfspId(debtorDfspId)
+  const creditorProxyId = await getCache().lookupProxyByDfspId(creditorDfspId)
```
geka-evk, line 24: maybe, we can use Promise.all?

test/integration-override/handlers/positions/handlerBatch.test.js
```diff
@@ -1838,6 +1839,10 @@ Test('Handlers test', async handlersTest => {
       await testConsumer.destroy() // this disconnects the consumers
 
       await Producer.disconnect()
+      try {
+        await ProxyCache.proxyCache.disconnect()
+      } catch (err) {
+      }
```
geka-evk, line 1845: maybe, we can add, at least, log with error here? (for easier debugging in case of error)

test/unit/api/ledgerAccountTypes/handler.test.js
```diff
@@ -37,6 +38,11 @@ Test('LedgerAccountTypes', ledgerAccountTypesHandlerTest => {
     sandbox = Sinon.createSandbox()
     sandbox.stub(Logger)
     sandbox.stub(LedgerAccountTypeService)
+    sandbox.stub(ProxyCache, 'getCache').returns({
+      connect: sandbox.stub(),
+      disconnect: sandbox.stub(),
+      healthCheck: sandbox.stub().returns(Promise.resolve())
```
geka-evk, line 44: the same, but with less code )
```
 ...
 healthCheck: sandbox.stub().resolves()
```

geka-evk (approved):

test/integration-override/handlers/positions/handlerBatch.test.js
kleyow: Steven refactored this in a follow up PR.

test/unit/api/ledgerAccountTypes/handler.test.js
kleyow: I'll clean this up in another PR. Copilot auto-complete makes me skip over sometimes.

kleyow: nvm. done. cleaned up here and throughout so copilot makes better suggestion next time.

src/lib/proxyCache.js
kleyow: done.

src/lib/proxyCache.js
geka-evk: I know that **ioredis** lib have [lazyConnect](https://redis.github.io/ioredis/interfaces/CommonRedisOptions.html#lazyConnect) option, which is _false_ by default (when a Redis instance is initialized, a connection to the server is immediately established). 
In our lib I set it by default to _true_: https://mojaloop.github.io/inter-scheme-proxy-cache-lib/types/RedisProxyCacheConfig.html#Default%20Value

So it's a bit weird why proxyCache is still trying to connect as soon as you instantiate the connection object

PR #1063, opened 2024-07-22
feat: fulfil obligation tracking

vijayg10:


src/handlers/transfers/handler.js
```diff
@@ -540,9 +540,15 @@ const processFulfilMessage = async (message, functionality, span) => {
           throw fspiopError
         }
       } else {
-        const payeeAccount = await Participant.getAccountByNameAndCurrency(transfer.payeeFsp, transfer.currency, Enum.Accounts.LedgerAccountType.POSITION)
-        await Kafka.proceed(Config.KAFKA_CONFIG, params, { consumerCommit, eventDetail, messageKey: payeeAccount.participantCurrencyId.toString(), topicNameOverride, hubName: Config.HUB_NAME })
-        histTimerEnd({ success: true, fspId: Config.INSTRUMENTATION_METRICS_LABELS.fspId })
+        const participantCurrencyId = await ProxyCache.deriveCurrencyId(transfer.payeeFsp, transfer.currency)
+        if (participantCurrencyId) {
+          await Kafka.proceed(Config.KAFKA_CONFIG, params, { consumerCommit, eventDetail, messageKey: participantCurrencyId.toString(), topicNameOverride, hubName: Config.HUB_NAME })
```
kleyow, line 545: Would you not need a `checkSameCreditorDebtorProxy` check here when publishing the messageKey to see if the key needs to be `0`?

vijayg10: You are right, I missed it. Its added now. Thanks

src/handlers/transfers/handler.js
```diff
@@ -197,88 +198,90 @@ const processFulfilMessage = async (message, functionality, span) => {
...
-          ]
-        }
-      }
-      message.value.content.payload = reservedAbortedPayload
-      await Kafka.proceed(Config.KAFKA_CONFIG, params, { consumerCommit, eventDetail: reserveAbortedEventDetail, fromSwitch: true, toDestination: transfer.payeeFsp, hubName: Config.HUB_NAME })
-    }
-
-    throw apiFSPIOPError
   }
+  // Lets validate FSPIOP Source & Destination Headers
```
kleyow, line 202: Why not use the headers to check if an fsp is inScheme and disabling these only if one is not in the scheme.
Feel like disabling these validatons entirely will break something in GP.

vijayg10: Thanks @kleyow addressed your concern.

PR #1072, opened 2024-07-25
feat(csi-334): updated deps

geka-evk:


vijayg10:
Closing the PR because deps are already updated as part of other commit

PR #1077, opened 2024-08-14
feat: impl fx abort

vijayg10:


src/domain/fx/cyril.js
```diff
@@ -203,12 +203,86 @@ const processFxFulfilMessage = async (commitRequestId) => {
...
+        notifyTo: transferRecord.payerFsp,
+        participantCurrencyId: transferPositionChange.participantCurrencyId,
+        amount: transferPositionChange.value
+      })
+    })
+  }
+  return positionChanges
+}
+
+const processFxAbortMessage = async (commitRequestId) => {
```
kleyow, line 243: nit: process doesn't seem like an correct descriptor of the function

vijayg10: I can't find a better name for this. I think you got this impression because its not doing much and just calculating the position changes to be done. But it aligns with the processFullFilMessage which does some more processing. That's why I followed the same naming convention for all the type of messages.

PR #1080, opened 2024-08-22
chore: add logging for admin api, proxy lookup, participant domain

kleyow:


src/lib/proxyCache.js
```diff
@@ -41,6 +43,7 @@ const getFSPProxy = async (dfspId) => {
 }
 
 const checkSameCreditorDebtorProxy = async (debtorDfspId, creditorDfspId) => {
+  logger.debug(`Checking if ${debtorDfspId} and ${creditorDfspId} are using the same proxy`)
```
geka-evk, line 46: I'd definitely recommend avoid using template strings as log message. The main reason is debugging and searching in logs. Let me try to explain my point:
1. When I found some logs (say, using Grafana), and I want to find appropriate place in code:
image
It's much easier and faster to do it if that log message is "static" string, and I don't need to remember that, say, 3rd word is "dynamic", and I need to search only by first 2 words.  In this particular log, in Grafana I'll have something like:
`Checking if 1234-3442345 and 90905-23432-..... are using the same proxy`, right?  So I need to search only by `Checking if `. And  I can get different matches (other places in code). It's not a huge issue, of course, but anyway.

So from my perspective, having log like:
```
logger.debug('Checking if debtor and creditor are using the same proxy', {  debtorDfspId, creditorDfspId })
```
will simplify such searching

2. Vice versa approach: when you want to filter logs to find exactly this line (using Grafana or grep). Again, I can search only by `Checking if`, which not exactly what I need.

3. If you use template string, you have to use `.isDebudEnabled` approach:
 ```
 logger.isDebugEnabled && logger.debug(`Checking if ${debtorDfspId} and ${creditorDfspId} are using the same proxy`)
 ```
 Personally, I don't think it affects performance much, but I had several "recommendation" about that from Vijay.
 
 As  a summary, I don't see any reason to use template string as log message, if we have possibility to pass second argument as an object with detailed information.

src/domain/participant/index.js
```diff
@@ -59,8 +60,10 @@ const { destroyParticipantEndpointByParticipantId } = require('../../models/part
 
 const create = async (payload) => {
   try {
+    logger.debug('creating participant with payload', payload)
     return ParticipantModel.create({ name: payload.name, isProxy: !!payload.isProxy })
   } catch (err) {
+    logger.error('error creating participant', { err, payload })
```
geka-evk, line 66: I'm not sure it will work, I assume we'll have in logs something like this:
```
2024-08-21T09:41:51.659Z - error: - error creating participant: - { "err": {} , payload: { // some data } }
```
for now in logger we can log error details only if it passed like this: logger.error('error message', err)

Si in this particular case, if we want to have `payload` be logged in every logs withing the function, we can use logger.child() approach:
```
   const log = logger.child(payload)  // or const log = logger.child({ payload }) to have payload-field 
   ...
   logger.debug('creating participant with payload')
   ...
   logger.error('error creating participant', err)
```

src/domain/participant/index.js
```diff
@@ -94,63 +101,76 @@ const getByName = async (name) => {
 }
 
 const participantExists = (participant, checkIsActive = false) => {
+  logger.debug('checking if participant exists', { participant, checkIsActive })
   if (participant) {
     if (!checkIsActive || participant.isActive) {
       return participant
     }
+    logger.error('participant is inactive', { participant })
```
geka-evk, line 109: maybe, `warn` severity is more appropriate here?

src/domain/participant/index.js
```diff
@@ -94,63 +101,76 @@ const getByName = async (name) => {
...
   try {
+    logger.debug('updating participant', { name, payload })
     const participant = await ParticipantModel.getByName(name)
     participantExists(participant)
     await ParticipantModel.update(participant, payload.isActive)
     participant.isActive = +payload.isActive
     participant.currencyList = await ParticipantCurrencyModel.getByParticipantId(participant.participantId)
     return participant
   } catch (err) {
+    logger.error('error updating participant', { name, payload, err })
```
geka-evk, line 126: https://github.com/mojaloop/central-ledger/pull/1080/files#r1726750337

src/domain/participant/index.js
```diff
@@ -94,63 +101,76 @@ const getByName = async (name) => {
...
   }
 }
 
 const createParticipantCurrency = async (participantId, currencyId, ledgerAccountTypeId, isActive = true) => {
   try {
+    logger.debug('creating participant currency', { participantId, currencyId, ledgerAccountTypeId, isActive })
     const participantCurrency = await ParticipantCurrencyModel.create(participantId, currencyId, ledgerAccountTypeId, isActive)
     return participantCurrency
   } catch (err) {
+    logger.error('error creating participant currency', { participantId, currencyId, ledgerAccountTypeId, isActive, err })
```
geka-evk, line 137: https://github.com/mojaloop/central-ledger/pull/1080/files#r1726750337

src/domain/participant/index.js
```diff
@@ -175,10 +195,13 @@ const destroyByName = async (name) => {
 
 const addEndpoint = async (name, payload) => {
   try {
+    logger.debug('adding endpoint', { name, payload })
     const participant = await ParticipantModel.getByName(name)
     participantExists(participant)
+    logger.debug('adding endpoint for participant', { participant, payload })
```
geka-evk, line 201: maybe, `info` severity?
  From my perspective, to define what severity to use  we should answer the question: Do we need to have this information during debugging production only by using logs?  If `yes` - then use `info`

kleyow: done, here and whereever appropriate.

src/domain/participant/index.js
kleyow: done

src/domain/participant/index.js
kleyow: done, here and throughout.

src/lib/proxyCache.js
kleyow: done

geka-evk (approved):

PR #1082, opened 2024-08-22
ci: use redis-cluster for integration tests by default

oderayi:


oderayi:
Moved to https://github.com/mojaloop/central-ledger/pull/1083

PR #1092, opened 2024-09-10
feat(csi-318): added externalParticipant table

geka-evk:


src/models/participant/externalParticipant.js
```diff
@@ -0,0 +1,118 @@
...
+
+// todo: think, if we need update method
+module.exports = {
+  create,
+  getOneByNameCached,
+  getOneByName,
+  getOneById,
+  destroyById,
+  destroyByName
+}
```
kleyow, line 118: Any strong reason moving away from other cache setups like `models/participant/participant.js` and `models/participant/participantCached.js`

geka-evk: No, I just decided to add cache a bit later, when all the rest logic is done

.nycrc.yml
```diff
@@ -28,6 +28,8 @@ exclude: [
   'src/handlers/transfers/FxFulfilService.js',
   'src/models/position/batch.js',
   'src/models/fxTransfer/**',
+  'src/models/participant/externalParticipantCached.js', # todo: figure out why it shows only 50% coverage in Branch
```
geka-evk, line 31: image

vijayg10 (approved):

PR #1119, opened 2024-11-04
feat: deprecated dependencies check

s-prak:
added deprecated dependencies check in circleci to check if any deprecated dependencies exist in the dependency tree of the repository


elnyry-sam-k:
@s-prak this can be closed now as it is part of the standard orb and ci config here

PR #1129, opened 2024-12-04
feat: propagate isOriginalId

oderayi:


src/domain/position/abort.js
```diff
@@ -157,6 +157,7 @@ const _constructAbortResultMessage = (binItem, id, from, notifyTo, isOriginalId,
     (isFx && !isOriginalId) ? Enum.Events.Event.Action.FX_ABORT : binItem.message?.value.metadata.event.action, // This will be replaced anyway in Kafka.produceGeneralMessage function
     state
   )
+  metadata.event.isOriginalId = isOriginalId
```
vijayg10, line 160: Can you move the location to context object? That is where we pass the additional info between services.

PR #1132, opened 2024-12-06
fix(CSI-984): fx transfer error duplicate check

oderayi:
- Refactored code slightly and update comments to aide readability
- Fix bug by passing in action to `fxFulfilService.getDuplicateCheckResult(...)`

### DIFF
--- a/src/handlers/transfers/handler.js
+++ b/src/handlers/transfers/handler.js
@@ -176,9 +176,9 @@ const processFulfilMessage = async (message, functionality, span) => {
   Logger.isInfoEnabled && Logger.info(Util.breadcrumb(location, { path: 'getById' }))
 
   const transfer = await TransferService.getById(transferId)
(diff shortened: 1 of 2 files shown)


src/handlers/transfers/handler.js
```diff
@@ -726,7 +730,7 @@ const processFxFulfilMessage = async (message, functionality, span) => {
     ['success', 'funcName']
   ).startTimer()
 
-  const dupCheckResult = await fxFulfilService.getDuplicateCheckResult({ commitRequestId, payload })
+  const dupCheckResult = await fxFulfilService.getDuplicateCheckResult({ commitRequestId, payload, action })
```
oderayi, line 733: This is the actual fix. `fxFulfilService.getDuplicateCheckResult` already implements correct duplicate error check based on `action`. Previously, `action` was not passed in.

vijayg10 (approved):

PR #1133, opened 2024-12-10
fix(csi-1016): fx abort fixes

oderayi:


src/domain/fx/cyril.js
```diff
@@ -250,16 +250,35 @@ const _getPositionChanges = async (commitRequestIdList, transferIdList, original
...
+    if (isAbort && transferRecord && transferRecord.payeeIsProxy && transferPositionChanges.length === 0 && !!commitRequestIdList.length && originalId === transferId) {
       positionChanges.push({
         isFxTransferStateChange: false,
         transferId,
         isOriginalId: originalId === transferId,
-        notifyTo: transferRecord.externalPayerName || transferRecord.payerFsp,
-        participantCurrencyId: transferPositionChange.participantCurrencyId,
-        amount: -transferPositionChange.change
+        notifyTo: transferRecord.payerFsp,
+        participantCurrencyId: transferRecord.payerParticipantCurrencyId,
```
vijayg10, line 267: Does it need to be null?

vijayg10: And is there any reason, you want to use the existing positionChanges array? I think its better to use a new field in context like `transferStateChanges` for this. Because positionChanges is used in the position handler to recursively change the positions by producing followup position messages. I am not sure if its a good idea to messup with this variable.

src/domain/position/abort.js
```diff
@@ -100,6 +100,13 @@ const processPositionAbortBin = async (
             resultMessages.push({ binItem, message: Utility.clone(resultMessage) })
           }
         }
+        // Add notifications in the transferStateChanges if available
+        if (Array.isArray(cyrilResult.transferStateChanges)) {
+          for (const transferStateChange of cyrilResult.transferStateChanges) {
+            const resultMessage = _constructAbortResultMessage(binItem, transferStateChange.transferId, from, transferStateChange.notifyTo, transferStateChange.isOriginalId, false)
+            resultMessages.push({ binItem, message: Utility.clone(resultMessage) })
+          }
+        }
```
vijayg10, line 109: Don't we need to change the transferstates in DB?

vijayg10 (approved):

PR #1160, opened 2025-03-18
feat: rethrow with context

kalinkrustev:


kleyow (approved):

src/api/ledgerAccountTypes/handler.js
```diff
@@ -30,7 +30,7 @@
 
 const LedgerAccountTypesService = require('../../domain/ledgerAccountTypes')
 const ErrorHandler = require('@mojaloop/central-services-error-handling')
-const { rethrow } = require('@mojaloop/central-services-shared').Util
+const rethrow = require('@mojaloop/central-services-shared').Util.rethrow.with('CL')
```
geka-evk, line 33: I believe, it's better to put such logic in one place, say `/src/shared/index.js`
```
const rethrow = require('@mojaloop/central-services-shared').Util.rethrow.with('CL')
...
module.exports = {
   rethrow,
   ...
}
```
and reuse it from there
```
const { rethrow } = require('./src/shared')
```

geka-evk (approved):

src/api/ledgerAccountTypes/handler.js
kalinkrustev: OK, will do this for the next PR

PR #1166, opened 2025-04-23
refactor: add a typescript build step to allow js and ts alongside each other

lewisdaly:
In preparation for the TigerBeetle integration, I want to introduce Typescript gradually into this project. 

I also added the file `demo.ts` as a sanity check to ensure that ts was compiling to js as expected.

Unfortunately since the `./migrations` and `./seeds` directories were being accessed outside of the root of `./src`, I had to move them inside of `./src` for `tsc` to pick them up and correctly emit them alongside the compiled `.ts` files into `./dist`. This does come with the benefit of simplifying the `Dockerfile` however.


lewisdaly:
Integration tests are failing, there seems to be some imports related to `proxyCache` that make the global variable `cacheClient` in `src/models/participant/participantCached.js` undefined. When I set `PROXY_CACHE.enabled=false`, that fixes those tests but breaks others.

PR #1170, opened 2025-06-02
chore(sbom): add initial sbom

s-prak:


elnyry-sam-k:
deps failed but also in the name use 19.5.1 maybe? corresponding to the current release #

PR #1171, opened 2025-06-02
chore: update check to account for false instead of just errors

kleyow:


src/lib/healthCheck/subServiceHealth.js
```diff
@@ -44,8 +44,22 @@ const MigrationLockModel = require('../../models/misc/migrationLock')
...
   let status = statusEnum.OK
+
   try {
-    await Promise.all(consumerTopics.map(t => Consumer.isConnected(t)))
+    const results = await Promise.all(
+      consumerTopics.map(async (t) => {
+        try {
+          return await Consumer.isConnected(t)
+        } catch (err) {
+          Logger.isDebugEnabled && Logger.debug(`isConnected threw for topic ${t}: ${err.message}`)
```
geka-evk, line 54: maybe, it's better to use `.warn` severity to log such errors?

geka-evk (dismissed):

src/lib/healthCheck/subServiceHealth.js
kleyow: done.

elnyry-sam-k (approved):
👍

PR #1188, opened 2025-07-09
feat: additional db options in config

vijayg10:


kirgene:
Continuing our discussion on PR (#1191), I found this related thread: https://github.com/knex/knex/issues/3204

Turns out the original `mysql` client actually returns decimals as numbers, so our previous code was already working with numbers by default.

If we disable `decimalNumbers: true` in `mysql2`, we'll start getting strings instead. And since we're not using TypeScript, we can't always be sure we're comparing values correctly. Even if tests pass, it might just be because `"10.00" < "20.00"` works as expected, but something like `"100.00" < "20.00"` would return `true`, which is clearly wrong.

So we could either:
- keep `decimalNumbers: true` to match the previous behavior, or
- switch right away to using `@mojaloop/ml-number` to handle parsing explicitly and safely across all handlers.

src/api/participants/handler.js
```diff
@@ -252,8 +253,8 @@ const getLimits = async function (request) {
           currency: (item.currencyId || request.query.currency),
           limit: {
             type: item.name,
-            value: item.value,
-            alarmPercentage: item.thresholdAlarmPercentage
+            value: new MLNumber(item.value).toNumber(),
+            alarmPercentage: item.thresholdAlarmPercentage !== undefined ? new MLNumber(item.thresholdAlarmPercentage).toNumber() : undefined
```
geka-evk, line 257: I'd suggest adding a simple function to avoid code duplication:
```
const toMLNumber = (value) => (value !== undefined ? new MLNumber(value).toNumber() : undefined)
````
and then in all 3 places:
```
...
alarmPercentage: toMLNumber(item.thresholdAlarmPercentage)

```

File: src/api/participants/handler.js [hunk 2]
```diff
@@ -308,15 +309,41 @@ const adjustLimits = async function (request, h) {
...
+      // Multiple positions (no currency specified)
+      return result.map(position => ({
+        ...position,
+        value: position.value !== undefined ? new MLNumber(position.value).toNumber() : undefined
+      }))
+    } else if (result && typeof result === 'object' && result.value !== undefined) {
+      // Single position (currency specified)
+      return {
+        ...result,
+        value: new MLNumber(result.value).toNumber()
```
geka-evk, line 325: What if `result.value`  if _undefined_ or _null_?

vijayg10: Good catch, just checked the MLNumber library and its handling these cases well enough. Thanks

config/knexfile.js
```diff
@@ -6,9 +6,8 @@ const seedsDirectory = '../seeds'
 const Config = require('../src/lib/config')
 
 module.exports = {
-  client: 'mysql2',
-  connection: Config.DATABASE.connection,
-  pool: Config.DATABASE.pool,
+  ...Config.DATABASE,
+  version: '5.5',
```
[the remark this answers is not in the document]
vijayg10: I am not sure why I added this. Its not needed, removed. Thanks

PR #1189, opened 2025-07-10
fix(knex): switch to mysql2 client and update version for mysql v8 auth compatibility

kirgene:
…th compatibility


elnyry-sam-k (dismissed):
👍

PR #1191, opened 2025-07-11
chore: use mysql2 dialect

kirgene:


elnyry-sam-k (approved):
👍

src/lib/config.js
```diff
@@ -44,7 +44,9 @@ module.exports = {
       port: RC.DATABASE.PORT,
       user: RC.DATABASE.USER,
       password: RC.DATABASE.PASSWORD,
-      database: RC.DATABASE.SCHEMA
+      database: RC.DATABASE.SCHEMA,
+      decimalNumbers: true,
```
vijayg10, line 48: Hi @kirgene I have another PR to switch mysql2 with additional options. Please review this https://github.com/mojaloop/central-ledger/pull/1188 and let me know if it makes sense.

Setting decimalNumbers in mysql2 may lead decimal number operation issues in javascript which we already have with mysql. Instead we can handle the parsing of decimal number in the models wherever needed in my opinion.
For example, in position calculations, we don't want db client lib convert the decimals from DB automatically, we want to handle that using ml-number lib for precise operations on the floating numbers in javascript.

So if my above PR is sufficient for your mysql2 change, we can merge that in instead of yours.

kirgene: Thanks @vijayg10 , I took a look at your PR, and one thing I noticed is that it uses `parseFloat()`, which is exactly what tends to cause floating-point issues in JS.

In my PR, I use `decimalNumbers: true` from mysql2, which mirrors the behavior of mysql. It parses decimals as floats under the hood, so it's backward-compatible and avoids changing existing logic.

Long term, the ideal way is to keep decimals as strings (which is the default in mysql2) and explicitly handle them with something like ml-number or decimal.js wherever precision matters. That gives us full control and avoids unexpected rounding.

So I think my PR offers a cleaner path for now

vijayg10: > In my PR, I use decimalNumbers: true from mysql2, which mirrors the behavior of mysql. It parses decimals as floats under the hood, so it's backward-compatible and avoids changing existing logic.

I don't think this is the case, please check this doc

PR #1207, opened 2025-09-05
feat(csi-1753): enhance distributed lock management with automatic extension and fix metrics

oderayi:
feat(csi-1752 & csi-1753): fix metrics and  enhance distributed lock management with automatic extension:
- Fix "metrics counter not initialized" error in timeout handler
- Implemented automatic lock extension for distributed locks, extending the lock at one third of the TTL.
- Added error handling for lock extension failures, ensuring the extension timer is cleared upon lock release.
- Updated logging to provide feedback on lock extension and release actions.

### DIFF
--- a/src/handlers/timeouts/handler.js
+++ b/src/handlers/timeouts/handler.js
@@ -374,7 +374,6 @@ const acquireLock = async () => {
 
         // Store the interval ID so we can clear it when the lock is released
         distLock.extensionTimer = lockExtender
-        // distLock.extensionCount is already initialized above
       }
       return acquired
     } catch (err) {


src/handlers/timeouts/handler.js
```diff
@@ -345,37 +348,70 @@ const initLock = async () => {
...
-      return !!(await distLock.acquire(distLockKey, distLockTtl, distLockAcquireTimeout))
+      const acquired = !!(await distLock.acquire(distLockKey, distLockTtl, distLockAcquireTimeout))
+      if (acquired) {
+        let extensionCount = 0
+
+        // Set up automatic lock extension at one third of the lock TTL
+        const extensionInterval = Math.floor(distLockTtl / 3)
+        const lockExtender = setInterval(async () => {
+          try {
+            await distLock.extend(distLockTtl)
```
geka-evk, line 366: shouldn't we use here `extensionInterval` instead of `distLockTtl`?

geka-evk (dismissed):

src/handlers/timeouts/handler.js
oderayi: No, we extend the lock by the original ttl (`distLockttl`) but we do that every 1/3 interval of the ttl (`extensionInterval`)

PR #1212, opened 2025-09-10
fix(csi-1748): use local lock with distlock

oderayi:


src/handlers/timeouts/handler.js
```diff
@@ -473,5 +477,6 @@ module.exports = {
   registerAllHandlers,
   registerTimeoutHandler,
   isRunning,
-  stop
+  stop,
+  running
```
kalinkrustev, line 481: This is not useful for anything, it just adds a property to the exports object, initialized with the default value of false.

kalinkrustev (approved):

PR #1279, opened 2026-03-12
fix: optimize bin processor to run multiple selects in a single DB request

kalinkrustev:


geka-evk (commented):
### Nit: Misleading query builder table

In `src/models/position/batch.js`, the extension query:
```js
knex('transfer')
  .transacting(trx)
  .from('transferExtension')
  .whereIn('transferId', reservedActionTransferIdList)
```

`knex('transfer')` initializes the query builder with the `transfer` table, then `.from('transferExtension')` overrides it. While `.from()` does override, this is confusing — readers expect `knex('tableName')` to be the actual table queried.

Should be:
```js
knex('transferExtension')
  .transacting(trx)
  .whereIn('transferId', reservedActionTransferIdList)
```

src/models/position/batch.js
```diff
@@ -263,7 +263,174 @@ const getReservedPositionChangesByCommitRequestIds = async (trx, commitRequestId
...
+          'tf.ilpFulfilment AS fulfilment',
+          'te.errorCode',
+          'te.errorDescription'
+        ),
+      // Pre fetch transfer extensions for all reserve action fulfils
+      reservedActionTransferIdList &&
+      reservedActionTransferIdList.length > 0 &&
+      knex('transfer')
+        .transacting(trx)
+        .from('transferExtension')
```
geka-evk, line 354: I'd suggest a small simplification: 
```js
knex('transferExtension')
  .transacting(trx)
  .whereIn('transferId', reservedActionTransferIdList) ...
```

src/models/position/batch.js
```diff
@@ -263,7 +263,174 @@ const getReservedPositionChangesByCommitRequestIds = async (trx, commitRequestId
...
+
+  // 4. Filter results (MySQL often adds a trailing header if the last query is an DML)
+  // This ensures you only get the data sets for your specific queries
+  return results.slice(0, queryBuilders.length)
+}
+
+const fetchAll = async (trx, transfersIdList, commitRequestIdList, accountIds, reservedActionTransferIdList) => {
+  const knex = Db.getKnex()
+  try {
+    const [results, results1, /* participantPositions */, transferInfos, participantPositionChanges, query, extensions] = await executeMultiQuery(trx, [
```
geka-evk: I think, we need to have something like this:
```
/* participantPositions, */
```
Move the comma inside the /* */, otherwise transferInfos, participantPositionChanges, query, extensions will get wrong dataset

src/lib/config.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
       database: RC.DATABASE.SCHEMA,
       ...RC.DATABASE.ADDITIONAL_CONNECTION_OPTIONS,
       decimalNumbers: false,
+      multipleStatements: true,
```
geka-evk: So now every Knex query in the app runs on a connection that accepts semicolon-separated multi-statement execution. So it might have security concern, that `knex.raw('...')` with unsanitized input becomes a multi-statement SQL injection vector, e.g.  `('; DROP TABLE transfer; --)`
Do we really need it?

src/models/position/batch.js
kalinkrustev: I am not sure, because of the select *, which will select from both tables. I have not checked if the consuming code needs the data from the transfer table.

src/lib/config.js
kalinkrustev: This is needed to run multiple queries together. The unsanitized input should not be allowed anyway. If needed, we can have a separate Knex for the batch queries. Unfortunately the performance always comes with a cost.

I recommended ~10 years ago to not use query builders for everything during one of the initial convenings, exactly due to things like this. A lot of overhead can be avoided by using stored procedures.

vijayg10 (approved):

PR #1285, opened 2026-03-31
fix: participant cache invalidation

vijayg10:


test/unit/lib/headerUtils.test.js
```diff
@@ -0,0 +1,84 @@
...
+ *   - Name Surname <name.surname@mojaloop.io>
+ *
+ * * Infitx
+ *   - Vijay <vijaya.guthi@infitx.com>
+ * --------------
+ */
+'use strict'
+const test = require('tape')
+const sinon = require('sinon')
+const Logger = require('../../shared/logger').logger
```
geka-evk: is it correct import path? I thought it should get logger from /src

test/unit/lib/headerUtils.test.js
```diff
@@ -0,0 +1,84 @@
...
+const {
+  getNormalizedHeaderValue,
+  parseBaggageHeader,
+  shouldSkipParticipantCache
```
geka-evk, line 36: **CRITICAL: Broken import paths** — these `require()` paths resolve to `test/shared/logger` and `test/lib/headerUtils` (neither exists). All sibling tests in `test/unit/lib/` use `../../../src/...`.

This test will crash on `require()`.

```suggestion
const Logger = require('../../../src/shared/logger').logger
const {
  getNormalizedHeaderValue,
  parseBaggageHeader,
  shouldSkipParticipantCache
} = require('../../../src/lib/headerUtils')
```

src/lib/proxyCache.js
```diff
@@ -83,9 +83,9 @@ const getCache = () => {
  * @param {Object} [options] - { validateCurrencyAccounts: boolean, accounts: [ { currency: string, accountType: Enum.Accounts.LedgerAccountType } ] }
  * @returns {ProxyOrParticipant} proxyOrParticipant details
  */
-const getFSPProxy = async (dfspId, options = null) => {
+const getFSPProxy = async (dfspId, options = null, skipParticipantCache = false) => {
```
geka-evk, line 87: **HIGH: JSDoc not updated** — `skipParticipantCache` param added to the signature but missing from the JSDoc above.

```suggestion
const getFSPProxy = async (dfspId, options = null, skipParticipantCache = false) => {
```

Add to JSDoc:
```js
 * @param {boolean} [skipParticipantCache=false] - Bypass participant cache lookup.
```

geka-evk (changes requested):
Review: 2 issues found (1 critical, 1 high).

Overall the extraction to `headerUtils.js` and the proxy cache bypass are good improvements. The two inline comments below need addressing before merge.

geka-evk (approved):

geka-evk (approved):

PR #1287, opened 2026-07-03
test: rewrite integration tests in Typescript, using new test harness

lewisdaly:
## Motivation

As we bring TigerBeetle into Mojaloop, I've found the current integration test setup (where the harness lives outside of the runner, dependent on docker, coordinated across a number of shell scripts) to be quite brittle and hard to debug.

Since we want to switch to the native nodejs test runner anyway, we decided it was best to bring this work in now instead of waiting before we bring in TigerBeetle. This improves the speed of local deveopment, and makes the integration tests on CI less brittle.

Additionally, this brings the time it takes to run the integration tests in CI from 17 minutes -> ~6 minutes, with more improvements to be made by increasing the parallelism and tightening timeouts later.

## How it Works

Each `*.int.ts` file is an integration test file, and is resonsible for running it's own infrastructure with `Harness`. (currently mysql, kafka, redis). Actually it's mariadb and redpanda since they give much better startup times.

For example:

```js
import Harness from './harness'
const harness = harness.getInstance()

// Then, in the before() and after() hooks: 
before(async () => {
  await harness.up()
  await harness.setupGlobals()
})

after(async () => {
  await harness.teardownGlobals()
  await harness.down()
})
```

Since each test file has it's isolated infrastructure, we don't have to code around existing database state, or run the tests in a given order to ensure the database has the expected state.

### Api Helpers
I've also introduced a `api-helpers.ts` file, which gives us higher level abstractions for doing common things inside of Mojaloop, with sensible defaults. For example, to create and prepare a payment, we can do:

```js
await ApiHelpers.buildPayment()
      .deps(harness, handlerAll)
      .parties('dfsp_a', 'dfsp_b')
      .transferId('1000001')
      .build()
      .prepare()
```

Internally, the harness tracks the kafka messages being produced and consumed, and waits until the required messages are produced and consumed, or fails if the incorrect number of messages were produced.

### Snapshot Testing

At TigerBeetle, [we use snapshot testing extensively](https://tigerbeetle.com/blog/2024-05-14-snapshot-testing-for-the-masses/) as it helps both easily test assertions in our tests, but also give additional context to the API contract to the reader. 

For example, it's much easier and more meaningful to the reader to write a snapshot assertion like so:

```js
const messages = harness.spoolLastPayload(1)
Snapshot.from(`[
  {
    "errorInformation": {
      "errorCode": "3106",
      "errorDescription": "Modified request",
      "extensionList": {
        "extension": [
          {
            "key": "cause",
            "value": :ignore
          }
        ]
      }
    }
  }
]`).checkUnwrap(messages)
```

Than the equivalent assertions:
```js
assert.equal(messages.length, 1)
assert(messages[0].errorInformation)
assert.equal(messages[0].errorInformation.errorCode, "3106")
assert.equal(messages[0].errorInformation.errorDescription, "Modified request")
assert(messages[0].errorInformation.extensionList)
assert(messages[0].errorInformation.extensionList.extension)
assert.equal(messages[0].errorInformation.extensionList.extension.length, 1)
assert.equal(messages[0].errorInformation.extensionList.extension[0].key, "cause")
assert(messages[0].errorInformation.extensionList.extension[0].key.value)
```

When we make changes to the API, we can then rerun the tests with `SNAP_UPDATE=1 node --test path/to/file`,  and the snapshot testing library will update the snapshot for us.

**AI Assistance Disclosure**
AI tools were used in producing part of this contribution.
Tools used: Claude Code
Scope: Claude was used to help analyse existing integration tests, and trace the expected test data passed to different components. All code was handwritten by @lewisdaly.
All AI-generated content has been reviewed, understood, and validated by the author.


test/scripts/test-integration.sh
```diff
@@ -50,14 +50,18 @@ bash .circleci/curl-retry-cl-health.sh
...
 ## Start integration tests
 echo "Running Integration Tests"
-LOG_LEVEL=${LOG_LEVEL} npm run test:xint
+# LOG_LEVEL=${LOG_LEVEL} npm run test:xint
+LOG_LEVEL=${LOG_LEVEL} npx tape 'test/integration/**/*.test.js'
 INTEGRATION_TEST_EXIT_CODE="$?"
 echo "==> integration tests exited with code: $INTEGRATION_TEST_EXIT_CODE"
 
+exit 0
+
```
elnyry-sam-k: Remove dead code below? Or clean up otherwise..

src/handlers/transfers/fx-abort.int.ts
```diff
@@ -0,0 +1,209 @@
...
+ * TigerBeetle
+ - Lewis Daly <lewis@tigerbeetle.com>
+ --------------
+ ******/
+
+import { after, before, describe, it } from "node:test"
+import Harness from '../../testing/harness'
+import * as ApiHelpers from '../../testing/api-helpers'
+import assert from "node:assert"
+import { assertPositionDiff, sleepSeconds } from "../../testing/util"
```
elnyry-sam-k: Sonarcloud suggestion:

> Remove this unused import of 'sleepSeconds'.

src/testing/harness.ts
```diff
@@ -956,8 +956,8 @@ class MySql {
     for (let attempt = 1; attempt <= attemptsMax; attempt++) {
       try {
         switch (this.options.migration.type) {
-          case "knex": return this.migrateKnex();
-          case "sql": return this.migrateSql();
+          case "knex": await this.migrateKnex();
```
elnyry-sam-k: end the switch case

src/domain/position/fx-timeout-reserved.js
```diff
@@ -167,6 +171,8 @@ const _constructFxTimeoutReservedResultMessage = (binItem, commitRequestId, coun
 }
 
 const _handleParticipantPositionChange = (runningPosition, transferAmount, commitRequestId, accumulatedPositionReservedValue) => {
+  assert(runningPosition !== undefined)
+  assert(transferAmount !== undefined)
```
elnyry-sam-k, line 175: There's already an explicit check above, still needed?

src/domain/position/fx-timeout-reserved.js
```diff
@@ -167,6 +171,8 @@ const _constructFxTimeoutReservedResultMessage = (binItem, commitRequestId, coun
 }
 
 const _handleParticipantPositionChange = (runningPosition, transferAmount, commitRequestId, accumulatedPositionReservedValue) => {
+  assert(runningPosition !== undefined)
```
elnyry-sam-k, line 174: message too, for assertion error?

elnyry-sam-k:
Regarding the `src/testing/harness.snapshot.sql` - any steps / docs or CI step needed for maintenance, check drift ?

src/domain/position/fx-timeout-reserved.js
lewisdaly: Yep, we [love assertions at TigerBeetle to catch programmer error](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md). I've been getting in the habit here of putting in assertions in the existing code wherever I run a test that fails and encounter a "... is undefined" error, since it means some invalid data was passed through into our function.

lewisdaly:
> Regarding the `src/testing/harness.snapshot.sql` - any steps / docs or CI step needed for maintenance, check drift ?

We prepend this to the top of the file:
>-- Note: This file was generated by ./testing/harness.ts.
>-- It is used to speed up the integration tests, but needs to be recreated every time a new
>-- migration is added. Refer to ./src/testing/harness.ts for instructions of how to update this
>-- snapshot.

The good thing as well, is that we run the knex migrations after restoring from snapshot anyway, so it's forgiving for new migrations, and will throw on migration drift. In the ideal world we wouldn't need it, but it takes 10+ seconds to run the migrations from scratch each time.

src/domain/position/fx-timeout-reserved.js
elnyry-sam-k: for the runningposition fine, but the transferAmount had an explicit check earlier, so asked. Thanks!

src/testing/harness.ts
```diff
@@ -0,0 +1,1153 @@
...
+      event: any,
+      'protocol.createdAt': number,
+      trace: any
+    }
+  }
+}
+
+/**
+ * Starts a redpanda and redpanda console docker container.
+ * Redpanda is much quicker to start up than Kafka, so is more suitable for this testing harness.
```
elnyry-sam-k, line 677: Redpanda is fine, but redpanda UI is not needed for all CI runs, isn't it? maybe needed only for debugging if the user needs it.

lewisdaly: Yeah it's pretty essential when debugging what messages got produced/consumed and to debug when the handlers aren't connecting. It costs us nothing in startup time as it starts much faster than the db.

elnyry-sam-k: I don't mean redpanda, but the UI with it (when the runs happen in CI)

PR #1288, opened 2026-07-16
fix: return thresholdAlarmPercentage from participant limit facade fo…

gibaros:
…r bignumber.js 11 [mojaloop/#4479]

- ml-number 11.4.5 (bignumber.js 11) throws on undefined instead of silent NaN
- getParticipantLimitByParticipantCurrencyLimit now returns thresholdAlarmPercentage (cached + SQL branches)
- test fixtures updated to the real facade/DB shape (binProcessor, fx-timeout-reserved)
- node 24.18.0, orb 2.1.3, dependency updates and vulnerability fixes


elnyry-sam-k (approved):
👍

PR #1293, opened 2026-08-23
refactor: split Transfer handlers into payment-prepare, payment-fulfil, forex-prepare, forex-fulfil

lewisdaly:
There is a lot of nested business logic in the Transfer handlers which make it difficult to reason about and extend. This PR introduces a new `DispatchTransferHandler`: A custom TransferHandler that lets us gradually route from the legacy transfer handler to the refactored split handlers.

We can switch between `JOINED` and `SPLIT` transfer handlers with the [new config option:](https://github.com/mojaloop/central-ledger/pull/1293/changes#diff-4370dddfda21e0995d1432a991f1bfbc3786f3ad031264cf074e684dc55b31deR83-R90)

```
  /**
   * @description Passed through to DispatchTransferHandler. When SPLIT (default), uses the new
   * split payment-prepare, payment-fulfil, forex-prepare and forex-fulfil handlers.
   * 
   * @default 'SPLIT'
   */
  HANDLERS_TRANSFER_DISPATCH_MODE: 'JOINED' | 'SPLIT',
```

The overarching goal for this section of the refactor is to simplify the transfer business logic, which will then let us more easily fuse together the position handlers, and then eventually wrap this all in the in the upcoming `Ledger` interface. There's a lot of todos and incompleted stuff here, all of which will be tidied up as I continue refactoring the fused position handlers.

Additionally, this PR has the starting point for the new Ledger Fuzzer, a testing harness to fuzz the new ledger implementations.


elnyry-sam-k:
hi @lewisdaly , thanks for the  PR..

Is the design for this documented somewhere, or a proposal, maybe a sequence diagram - which can be used as reference while reviewing this.

In addition, while the other model is still in place, not sure if having an in-progress design change (with to-dos) get into main; does it make sense to have a separate branch for tigerbeetle here in ML github at this stage? There are still other changes happening on central-ledger, so I'm wondering if you'd be ok with such an option. Maybe we can discuss this on the evolution WS call today. Thanks!

lewisdaly:
As discussed in the evolution workstream call, we're going to take this back into another separate branch for now, since it's still a lot of heavy refactoring that needs to be done to be able to implement the new Ledger interface.

PR #1296, opened 2026-09-02
tracking: message layer and handlers refactor

lewisdaly:
This is a PR to track the overall changes for this refactor. I mainly want to make sure CI is green, and see the integration test speedup.


lewisdaly:
We no longer need this!

PR #1301, opened 2026-09-15
feat: commit kafka messages asynchronously instead of blocking commit

shashi165:
## Summary

- Bumps `@mojaloop/central-services-shared` to `18.39.3` and `@mojaloop/central-services-stream` to `11.10.0`, enabling the opt-in async Kafka commit strategy (`commitStrategy: 'async'`) used to eliminate per-message blocking commits on the hot path.
- Fixes a breaking change introduced by that bump: `central-services-shared`'s `apiDocumentation` Hapi plugin renamed its option from `documentPath` to `pathToSwaggerFile`. Updated both registration sites (`src/shared/plugins.js`, `src/settlement/shared/plugins.js`).
- Fixes the local pre-commit `test` hook crashing with `EPIPE` (`src/testing/run.ts`): the unit-test runner was relaying two raw, concatenated TAP streams (legacy tape + native `node:test`) straight to stdout instead of only the already-merged, flattened output, which broke `tap-spec`'s TAP parser; separately, `process.exit()` was called immediately after the final write, which can truncate output when stdout is piped (not a TTY). Both are fixed.
- Two small pre-existing bug fixes carried in this branch (not part of the dependency bump): a mislabeled histogram timer in `facade.js` that stopped before the DB commit actually completed, and two missing `await`s in `transferDuplicateCheck.js`.

## Test plan
- [x] Full unit suite passes: `npm run test:unit:spec` (3558/3558)
- [x] `npm run audit:check` passes
- [x] `npm run lint` clean on touched files
- [x] Verified async commit strategy end-to-end against real released `central-services-shared@18.39.3` / `central-services-stream@11.10.0`

---
## AI use disclaimer

This change was developed with AI assistance (Claude Code Sonnet 5). All changes were reviewed, tested, and verified by a human before submission.


gibaros (approved):
👍

elnyry-sam-k:
Thanks @shashi165 

one clarification - Is the `commitStrategy` itself set in helm or elsewhere, because there's no direct reference or config added here (or a default set appropriately in cs-stream)..

elnyry-sam-k (approved):
👍

shashi165:
> Thanks @shashi165
> 
> one clarification - Is the `commitStrategy` itself set in helm or elsewhere, because there's no direct reference or config added here (or a default set appropriately in cs-stream)..

Yes, it has to be set in helm values and it will be rendered in the `default.json`

PR #1302, opened 2026-09-16
fix: knexfile hardcoded mysql driver breaks migrations on MySQL 9

shashi165:
## Summary

Regression from #1243 ("feat: add typescript and basic `tsc` build"), which replaced `config/knexfile.js` with `src/knexfile.js` during the migrations/seeds directory move. The old file spread the whole `Config.DATABASE` object (`...Config.DATABASE`), so `client` came from config. The rewrite replaced that with explicit fields and hardcoded `client: 'mysql'` (plus a `version: '5.5'` that never existed before), silently dropping the config-driven driver selection.

`config/default.json` already correctly sets `DATABASE.DIALECT: "mysql2"`, and `src/lib/config/resolver.ts` already correctly maps that to `Config.DATABASE.client` — the knexfile just stopped reading it. Because of the hardcoding, knex actually loads and uses the legacy `mysql` driver for every migration run regardless of config, and that driver fails against MySQL 9 (its connection/auth protocol isn't supported). This isn't a theoretical incompatibility — it's the active driver behind `npm run migrate:latest`, `migrate:rollback`, and `seed:run`, and it fails outright against a MySQL 9 server even though `DATABASE_DIALECT` is already configured for `mysql2`.

## Changes

- `src/knexfile.js`: use `Config.DATABASE.client` instead of the hardcoded `'mysql'` literal, so the knexfile actually uses the configured `mysql2` driver instead of silently overriding it back to `mysql`. Dropped the unused hardcoded `version: '5.5'` alongside it.
- `package.json`: added `mysql2` to `optionalDependencies`. It was only resolvable by accident — present in `node_modules` solely as an undeclared transitive peer dependency of `@mojaloop/central-services-shared`, never declared by this project itself. Kept the existing `mysql` entry alongside it, since `DATABASE_DIALECT` is a free-form config string with no validation, so an environment could still legitimately configure `mysql`.
- `package.json`: added a `postbuild` script (`cp src/migrations/migration.template dist/migrations/migration.template`). Separate but adjacent gap found while verifying the fix: `tsc`'s `allowJs` compile step only emits `.js`/`.ts` files into `dist/`, so the plain-text migration stub template never made it into the build output. `dist/knexfile.js` (the file actually used by the `migrate:*` npm scripts, via `--knexfile ./dist/knexfile.js`) references this stub by a `dist`-relative path, so `npm run migrate:create` was broken independent of the driver issue.


gibaros (dismissed):
👍

package.json
```diff
@@ -36,6 +36,7 @@
   ],
   "scripts": {
     "build": "tsc",
+    "postbuild": "cp src/migrations/migration.template dist/migrations/migration.template",
```
elnyry-sam-k:
```suggestion
    "postbuild": "mkdir -p dist/migrations && cp src/migrations/migration.template dist/migrations/migration.template"
```

elnyry-sam-k (dismissed):
👍

elnyry-sam-k (approved):
👍

